PR #38, opened 2023-10-19
SUP-1487: GitHub Actions docs

james2791:
GitHub Action workflow migration docs

Added categories based on the most top-level keys definable in a workflow (i.e supported, partially, not-supported)


docs/GHA.md
```diff
@@ -0,0 +1,50 @@
...
+
+### on
+
+The `on` key allows for triggering a GitHub Action workflow. In Buildkite pipelines - this capability is defined within a `trigger` [step](https://buildkite.com/docs/pipelines/trigger-step) - where utilized within a pipeline, will create a build on the specified pipeline with additional properties.
+
+### permissions
+
+Within GitHub Action workflows, a `permissions` key allows the permission of the `GITHUB_TOKEN` to be modified with additional or removed access as required. Within Buildkite - [API Access Tokens](https://buildkite.com/docs/apis/managing-api-tokens) can be used within the context of a pipelines' build to interact with various Buildkite resources such as pipelines, artifacts, users, Test suites and more. Each token has a specified [token scope](https://buildkite.com/docs/apis/managing-api-tokens#token-scopes) that applies to interactions with the [REST](https://buildkite.com/docs/apis/rest-api) API, and can be configured with permission to interact with Buildkite's [GraphQL](https://buildkite.com/docs/apis/graphql-api) API.
```
toote, line 46: I would also clarify here that status check updates on Github are done through the application that must be connected and authorized for the integration to work and is not available for pipelines to use

docs/GHA.md
```diff
@@ -0,0 +1,50 @@
...
+
+### run_name
+
+This key sets the name of a GitHub Action workflow's run. Within Buildkite - when pipeline builds are created - a build message can be specified (contained in each job's environment as `BUILDKITE_MESSAGE`). The value is empty when a message is not set - and on events from source control, is the commit's title. A build message can also be set when creating a build through the [REST](https://buildkite.com/docs/apis/rest-api/builds#create-a-build) and [GraphQL](https://buildkite.com/docs/apis/graphql/schemas/mutation/buildcreate) APIs.
```
toote, line 50: I would also clarify that this configuration is specific to GitHub's UI and that is why there is no direct translation

docs/GHA.md
```diff
@@ -0,0 +1,50 @@
...
+### jobs
+
+GitHub Action workflows allows you to specify one or more `jobs` - main tasks of a workflow that run in parallel by default.
+
+The Buildkite Migration tool currently supports the following 
+
+- `runs_on`: The `runs_on` key defines the type of machine that the job will run on. Within Buildkite, this is mapped to an agent targeting [tag](https://buildkite.com/docs/agent/v3/queues#targeting-a-queue) of `runs_on`. Note that jobs that target custom `tag` names will have a `queue` target of `default`.
+- `steps`: Steps that are defined for a particular `job`. Any `run` key is supported - and at present, any action that defines a `uses` attribute currently is not supported.
+
+## Unsupported properties
```
toote, line 28: I would split these in 2 sections, those that are not supported because of lack of functionality (defaults) and those that are have no translation (name, on, permissions)

docs/GHA.md
```diff
@@ -0,0 +1,50 @@
...
+
+The `cancel-in-progress` Boolean value that can be defined in a `concurrency` hash inside a GitHub Action job workflow maps to the Buildkite pipeline setting of [Cancel Intermediate Builds](https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds) - which can be set within a pipeline's settings page, or when creating/updating a pipeline via the [REST](https://buildkite.com/docs/apis/rest-api/pipelines#create-a-yaml-pipeline)/[GraphQL](https://buildkite.com/docs/apis/graphql/schemas/mutation/pipelinecreate) APIs.
+
+### env 
+
+GitHub Actions allows the specification of environment variables at the workflow level, at a `job` level, and for each `step` within a `job`.
+
+Environment variables that are defined at the top of a workflow will be transition to [build level](https://buildkite.com/docs/pipelines/environment-variables#environment-variable-precedence) environment variables in the generated Buildkite pipeline. Environment variables defined within the context of each of a workflows' `jobs` are transitioned to [step level](https://buildkite.com/docs/pipelines/environment-variables#runtime-variable-interpolation) environment variables.
```
toote, line 17: why is this _partially supported_? It sounds like they are translated correctly

docs/GHA.md
```diff
@@ -0,0 +1,50 @@
...
+- `steps`: Steps that are defined for a particular `job`. Any `run` key is supported - and at present, any action that defines a `uses` attribute currently is not supported.
+
+## Unsupported properties
+
+### defaults
+
+A `defaults` key allows the definition of a hash in which configuration will apply to all of a workflow's jobs. 
+
+While there isn't a direct transition from a GitHub Action workflow to a Buildkite pipeline - the Buildkite platform allows jobs to have default configuration applied to them through the use of [Agent](https://buildkite.com/docs/agent/v3/hooks#agent-lifecycle-hooks) or [Job](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) Lifecycle hooks. Agent lifecycle hooks allows the configuration of customized commands on agent start up and shut down. Job Lifecycle hooks allow for the customization of what occurs at each stage of a job's lifecycle - and can be specified and applied at an agent, repository or plugin level.
```
toote, line 34: I'm not sure that this should be on the _unsupported_ list as there is a [partial (and untested) implementation of `defaults` ](https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/parsers/gha.rb#L41-L43)

That said, I am not sure why you are talking about hooks (unless it is to set environment variables.

docs/GHA.md
```diff
@@ -0,0 +1,50 @@
+# GitHub Actions
+
+The Buildkite Migration tool supports transitioning GitHub Action workflow definitions to Buildkite pipelines.
```
toote, line 3: This is specified in the readme, I would change this sentence to be more an introduction of what to expect from this document (more details about what kind of things they should not expect to be translated)

docs/GHA.md
mcncl: `workflow's`, rather than` workflows'`; a workflow is a singular.

docs/GHA.md
mcncl: I'm not sure that a default is the same as a hook? It feels more like YAML anchors than hooks?

mcncl (changes requested):

docs/GHA.md
james2791: Tweaked in the latest commit 👍 (added a GitHub App note / installation needed)

docs/GHA.md
james2791: Tweaked with the latest!

docs/GHA.md
james2791: Adjusted with both - I've added YAML anchors as the primary alternative - hooks below

docs/GHA.md
james2791: I moved it up to supported: I think I was missing the step level env / variable assignment (along with job level/build level)

docs/GHA.md
james2791: Split

docs/GHA.md
james2791: Added a snippet describing the GitHub App needs to be installed first for status checks (pertinent to the `permissions` allowing for the interaction of `statuses`

docs/GHA.md
james2791: This also is now partially supported (instead of being under unsupported)

mcncl (approved):

PR #46, opened 2023-10-23
Add docker login mapping

jradtilbrook:


toote (changes requested):
If we are going to start translating actions, I believe they should be part of [the corresponding step translation](https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/parsers/gha/steps.rb#L34-L36) to avoid the comments about it being unsupported

jradtilbrook:
Sweet thanks @mcncl. Just need to run `UPDATE_SNAPSHOTS=true bundle exec rspec`

jradtilbrook:
> Sweet thanks @mcncl. Just need to run `UPDATE_SNAPSHOTS=true bundle exec rspec`

Done

mcncl (approved):

PR #47, opened 2023-10-23
Implement actions/setup-python in step parser

mcncl:
CleanShot 2023-10-24 at 11 38 16

### DIFF
--- a/app/lib/bk/compat/parsers/gha/steps.rb
+++ b/app/lib/bk/compat/parsers/gha/steps.rb
@@ -32,7 +32,19 @@ module BK
       end
 
       def translate_uses(step)
-        "# action #{step['uses']} can not be translated just yet"
+        case step['uses']
+        when /\Aactions\/setup-python@v\d+\z/
+          python_version = step.dig('with', 'python-version') || 'latest'
+          { 
+            "plugins" => [
+              { 
+                "docker#v5.9.0" => {
+                  "image" => "python:#{python_version}"
+                }
+              }
+            ]
+          }
+        end
       end
 
       def generate_command_string(commands: [], env: {}, workdir: nil)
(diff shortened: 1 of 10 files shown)


app/lib/bk/compat/parsers/gha/steps.rb
```diff
@@ -32,7 +32,17 @@ def translate_run(step)
...
+        when /\Aactions\/setup-python@v\d+\z/
+          python_version = step.dig('with', 'python-version') || 'latest'
+          image_string = "python: #{python_version}"
+          BK::Compat::Plugin.new(
+            name: 'docker',
+            config: {
+              'image' => image_string
+            }
+          )
+        end
```
jradtilbrook, line 45: Going off the errors in the build from the snapshots, I think this might need an `else nil` type deal. and then an update to the snapshots and its looking good i think

jradtilbrook (changes requested):

app/lib/bk/compat/parsers/gha/steps.rb
mcncl: Good call! Added that support in.

mcncl: Ah, when is has a `uses` that doesn't match it inserts an empty line...

mcncl: Implemented a fix for that but need to assess the output; if we want it to match out "expected" then it'll need some work.

jradtilbrook: I pushed a change to fix the error and updated the snapshots. @mcncl i think its good now 👍

jradtilbrook (approved):

jradtilbrook:
The output snapshot with docker plugin is looking pretty good: https://buildkite.com/namekite/migration/builds/4#018b5f6b-c9c6-4d04-855b-335874320c59 its failing since this isnt a python project but thats expected

james2791 (approved):
👍 👍

PR #50, opened 2023-10-25
SUP-1518 Run snapshots on buildkite

jradtilbrook:


app/examples/github-actions/laravel-tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [8.1, 8.2, 8.3]
+        php: ['8.1', '8.2', '8.3']
```
toote, line 42: What is the reason for this change? If it is that the generated BK pipeline needs them to be strings, we should have a bugfix on the translation to ensure they are always strings instead of changing them here as this would still be a valid GHA configuration we might find

jradtilbrook: just a temporary fix for now to get the build green, ill work on ensuring its all strings soon

pzeballos: I don't think is a good idea to change tests in order to have green builds, that could bring lots of problems, not to mention that it is bad practice. The bugfix doesn't seem to be very complicated 🤔

jradtilbrook: hmm so a problem I'm having is that I can map the values to strings. But when its printed back to yaml, it seems to convert it back to numeric values again. I'm not sure why thats happening. See the latest commit for the mapping, but the output is still numeric

app/lib/bk/compat/parsers/gha/matrix.rb
```diff
@@ -5,6 +5,7 @@ module Compat
     class GitHubActions
       def set_matrix(bk_step, config)
         matrix = config['strategy']['matrix']
+        matrix.map { |key, value| [key, value.map(&:to_s)] }.to_h
```
toote, line 8: `map` returns a copy of the array, you can ensure all values are strings with:

```suggestion
        matrix = config['strategy']['matrix'].transform_values(&:to_s)
```

app/examples/github-actions/laravel-tests.yml
toote: left a comment with an explanation and way to solve the issue

app/lib/bk/compat/parsers/gha/matrix.rb
jradtilbrook: Oh damn that has changed the whole array to a string so its not quite what we want

toote: Oh... in that case I believe the solution would be have something like [the `recurse_to_string` method we use for transforming steps](https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/pipeline/step.rb#L169-L181) applied here

jradtilbrook:
Alright, got this working with some pairing with @toote

app/lib/bk/compat/parsers/gha/matrix.rb
```diff
@@ -1,15 +1,31 @@
...
+        case value
+        when Hash
+          value.transform_values! { |elem| convert_to_string(elem) }
+        when Array
+          value.map! { |elem| convert_to_string(elem) }
+        when Numeric
+          "#{value}"
+        else
+          # if we don't know how to do this, do nothing
+          value
```
toote, line 26: why not just?
```suggestion
        else
          value.to_s
```

toote (approved):

PR #55, opened 2023-10-27
SUP-1503 Add support for step-level branch filtering

lizrabuya:
This PR adds step-level branch filtering for the [on.push ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore) event workflow and supports `branches` and `branches-ignore`

### DIFF
--- a/app/lib/bk/compat/parsers/gha.rb
+++ b/app/lib/bk/compat/parsers/gha.rb
@@ -4,7 +4,7 @@ require_relative '../error'
 require_relative '../translator'
 require_relative '../pipeline'
 require_relative 'gha/jobs'
-require_relative 'gha/steps'
+require_relative 'gha/steps' 
 
 module BK
   module Compat
@@ -22,7 +22,7 @@ module BK
       end
 
       def self.matches?(text)
-        config = YAML.safe_load(text)
+        config = YAML.safe_load(text)      
         # YAML translates the `on` key in YAML to ruby's `true` boolean
         mandatory_keys = ['jobs', true]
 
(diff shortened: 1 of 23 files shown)


app/lib/bk/compat/parsers/gha.rb
```diff
@@ -4,7 +4,7 @@
 require_relative '../translator'
 require_relative '../pipeline'
 require_relative 'gha/jobs'
-require_relative 'gha/steps'
+require_relative 'gha/steps'
```
mcncl, line 7: Additional white space can be removed.

app/lib/bk/compat/parsers/gha.rb
```diff
@@ -22,7 +22,7 @@ def self.option
       end
 
       def self.matches?(text)
-        config = YAML.safe_load(text)
+        config = YAML.safe_load(text)
```
mcncl, line 25: Same here.

app/spec/lib/bk/compat/__snapshots__/examples/github-actions/basic.yaml.snap
```diff
@@ -7,3 +7,4 @@ steps:
     runs-on: ubuntu-latest
   label: ":github: explore"
   key: explore
+  branches: ''
```
mcncl, line 10: It doesn't seem to be handling null? Shouldn't we _not_ have the filter if one isn't set?

mcncl (changes requested):

app/spec/lib/bk/compat/__snapshots__/examples/github-actions/basic.yaml.snap
lizrabuya: Good catch. I did try different scenarios and this has regressed in the last changes I made.

lizrabuya: @mcncl , fixed in latest commit

mcncl (approved):

mcncl (approved):

PR #56, opened 2023-10-27
Map secrets straight through to env vars

jradtilbrook:


app/spec/lib/bk/compat/__snapshots__/examples/github-actions/java-spring-boot-bmi.yaml.snap
```diff
@@ -35,14 +35,13 @@ steps:
...
-  - touch key.txt && echo "Context element secrets.SERVER_KEY can not be translated
-    (yet)" > key.txt
+  - touch key.txt && echo "$$SERVER_KEY" > key.txt
   - chmod 600 key.txt
   - echo '~~~ Deploy the application'
   - "("
-  - SERVER_HOST="Context element secrets.SERVER_HOST can not be translated (yet)"
-  - SERVER_PORT="Context element secrets.SERVER_PORT can not be translated (yet)"
-  - SERVER_USER="Context element secrets.SERVER_USER can not be translated (yet)"
+  - SERVER_HOST="$$SERVER_HOST"
```
jradtilbrook, line 42: This looks weird and is redundant. I think we could fix that in a separate area to just not include the env vars and the subshell. CC @toote

app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -138,6 +138,11 @@ def self.replace_context_vars(var_name)
         "$#{var_name}"
       end
 
+    def self.replace_context_secrets(var_name)
```
james2791, line 141: Small 2col indent for this, otherwise quick and easy 👍

james2791 (changes requested):

app/lib/bk/compat/parsers/gha/expressions.rb
jradtilbrook: oh i think there mustn't be consistency in this project or something 🤷‍♂️ 
my editor indents things all wonky all the time

james2791 (approved):
🛫

app/spec/lib/bk/compat/__snapshots__/examples/github-actions/java-spring-boot-bmi.yaml.snap
toote: I agree it looks _dirty_, but it is an artifact of the small modular process of translating the original information as we never know where the variable is coming from (step commands are translated in one moment and contexts are translated in an isolated fashion - with no information as to where it is being translated on - at a later stage). It may be an interesting improvement, but a **very big** overhaul of the whole translation process.

Also note that secrets will also need manual intervention to setup plugins (in the step or the agent) to get those secrets available in the first place.

toote (approved):
While I'm OK with the code as it is, it may be a good idea to change the variable names being replaced to `SECRET_X` to make them easier to spot and/or search and replace for users afterwards

jradtilbrook:
> While I'm OK with the code as it is, it may be a good idea to change the variable names being replaced to `SECRET_X` to make them easier to spot and/or search and replace for users afterwards

Oh I like that. I'll change it now

app/spec/lib/bk/compat/__snapshots__/examples/github-actions/java-spring-boot-bmi.yaml.snap
jradtilbrook: Yeah okay makes sense

PR #57, opened 2023-10-27
SUP-1522: Supported GitHub context variable replacement to `BUILDKITE_x` variables

james2791:
Maps `${{ github.x }}` contexts to supported and relevant `BUILDKITE_y` variables when encountered in a passed through action YAML. 

Context name | Buildkite Environment Variable
--- | --- | 
`actor` | `BUILDKITE_BUILD_AUTHOR` |
`job` |`BUILDKITE_JOB_ID` |
`ref_name` | `BUILDKITE_BRANCH` |
`run_attempt` | `BUILDKITE_RETRY_COUNT` |
`run_id` | `BUILDKITE_BUILD_ID` |
`run_number` | `BUILDKITE_BUILD_NUMBER` |
`sha` | `BUILDKITE_COMMIT` |
`triggering_actor` | `BUILDKITE_BUILD_CREATOR` |
`workflow` | `BUILDKITE_PIPELINE_NAME` |

### DIFF
--- a/app/lib/bk/compat/parsers/gha/expressions.rb
+++ b/app/lib/bk/compat/parsers/gha/expressions.rb
@@ -122,7 +122,6 @@ module BK
 
     class GithubContext
       @github_bk_context_mapping = {
-        "action" => "BUILDKITE_PIPELINE_NAME",
         "actor" => "BUILDKITE_BUILD_AUTHOR",
         "job" => "BUILDKITE_JOB_ID",
         "ref_name" => "BUILDKITE_BRANCH",
--- a/app/lib/bk/compat/parsers/gha/expressions.rb
+++ b/app/lib/bk/compat/parsers/gha/expressions.rb
@@ -188,13 +188,9 @@ module BK
       end
 
       def self.replace_context_github(var_name)  
-        translated_var = @github_bk_context_mapping[var_name]
-        case translated_var
-        when String
-          "$#{translated_var}"
-        when nil
-          raise NoMethodError
-        end
+        raise NoMethodError unless @github_bk_context_mapping.include?(var_name)
+        
+        "$$#{@github_bk_context_mapping[var_name]}"
       end
 
       def self.replace_context_needs_outputs(job, path_list)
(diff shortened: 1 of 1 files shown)


app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -154,6 +166,16 @@ def self.replace_context_needs(path)
         end
       end
 
+      def self.replace_context_github(var_name)  
+        translated_var = @github_bk_context_mapping[var_name]
+        case translated_var
+        when String
+          "$#{translated_var}"
+        when nil
+          raise NoMethodError
```
jradtilbrook, line 175: I think this should instead print a value reporting to the user that it is not supported instead of raise an error

toote, line 175: I believe this could be changed to not use the `case` statement:
```suggestion
        raise NoMethodError unless @github_bk_context_mapping.include?(var_name)
        
        "$$#{@github_bk_context_mapping[var_name]}"
```

app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -154,6 +166,16 @@ def self.replace_context_needs(path)
         end
       end
 
+      def self.replace_context_github(var_name)  
+        translated_var = @github_bk_context_mapping[var_name]
+        case translated_var
+        when String
+          "$#{translated_var}"
```
toote, line 173: based on previous reviews and implementations this is missing a `$` to avoid escaping on pipeline upload:
```suggestion
          "$$#{translated_var}"
```

app/lib/bk/compat/parsers/gha/expressions.rb
toote: the `NoMethodError` exception is caught in the calling method and prints out the error we currently see saying that it can not be translated. it is an interesting idea but I probably agree that we should customize the error message

app/lib/bk/compat/parsers/gha/expressions.rb
james2791: Switched to the `raise``return method vs the case as above - including the dual `$$` 👆

app/lib/bk/compat/parsers/gha/expressions.rb
james2791: This idea is better, switched to this

toote (approved):

app/lib/bk/compat/parsers/gha/expressions.rb
james2791: Counted for this as below 👇

PR #58, opened 2023-10-27
Add runner context parsing

jradtilbrook:
This is best effort. Some things have mappings, some dont have direct mappings

### DIFF
--- a/app/lib/bk/compat/parsers/gha/expressions.rb
+++ b/app/lib/bk/compat/parsers/gha/expressions.rb
@@ -143,6 +143,22 @@ module BK
         "$$GITHUB_SECRET_#{var_name.upcase}"
       end
 
+      def self.replace_context_runner(var_name)
+        case var_name
+        when "name"
+          "$$BUILDKITE_AGENT_NAME"
+        when "os"
+          "$$BUILDKITE_AGENT_META_DATA_OS"
+        when "arch"
+          "$$BUILDKITE_AGENT_META_DATA_ARCHITECTURE"
+        when "temp"
+        when "tool_cache"
+          "$$TMPDIR"
+        else
+          ""
+        end
+      end
+
       def self.replace_context_needs(path)
         job, context, *rest = path.split('.')
         case context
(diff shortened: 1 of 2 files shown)


james2791 (approved):

app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -143,6 +143,21 @@ def self.replace_context_secrets(var_name)
...
+        when "name"
+          "$$BUILDKITE_AGENT_NAME"
+        when "os"
+          "$$BUILDKITE_AGENT_META_DATA_OS"
+        when "arch"
+          "$$BUILDKITE_AGENT_META_DATA_ARCHITECTURE"
+        when /(temp|tool_cache)/
+          "$$TMPDIR"
+        else
+          ""
```
toote, line 157: Previously it printed out a message stating it was not possible to translate. I'm not sure it is a good idea to just remove information that was previously present. How about doing something similar now?

```suggestion
          "There is no translation for github.#{var_name}"
```

app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -143,6 +143,21 @@ def self.replace_context_secrets(var_name)
...
 
+      def self.replace_context_runner(var_name)
+        case var_name
+        when "name"
+          "$$BUILDKITE_AGENT_NAME"
+        when "os"
+          "$$BUILDKITE_AGENT_META_DATA_OS"
+        when "arch"
+          "$$BUILDKITE_AGENT_META_DATA_ARCHITECTURE"
+        when /(temp|tool_cache)/
```
toote, line 154: according to the documentation, `tool_cache` is like `/usr/local/bin` or something like that. Not sure if mapping it to a tempdir is a good idea when there is no direct equivalent as agent environments and available executables are not uniform

app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -143,6 +143,21 @@ def self.replace_context_secrets(var_name)
         "$$GITHUB_SECRET_#{var_name.upcase}"
       end
 
+      def self.replace_context_runner(var_name)
+        case var_name
+        when "name"
```
toote, line 148: minor comment: non-interpolating strings should use single quotes (that is what running `rubocop app/` states and I believe it is because it avoids trying to parse and replace stuff in them)

toote (approved):

james2791 (approved):

PR #60, opened 2023-10-27
Support Matrix conversion

mcncl:
Marking this as ready just to get some input on what's not working here. It seems to be getting _something_ but I cannot test locally at the moment so doing so in CI.

### DIFF
--- a/app/Gemfile.lock
+++ b/app/Gemfile.lock
@@ -1,24 +1,13 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    ast (2.4.2)
     awesome_print (1.9.2)
     diff-lcs (1.5.0)
-    json (2.6.3)
-    language_server-protocol (3.17.0.3)
-    parallel (1.23.0)
-    parser (3.2.2.4)
-      ast (~> 2.4.1)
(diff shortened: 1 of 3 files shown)


app/lib/bk/compat/parsers/gha/expressions.rb
```diff
@@ -138,6 +138,10 @@ def self.replace_context_vars(var_name)
         "$#{var_name}"
       end
 
+      def self.replace_context_matrix(var_name)
+        var_name.gsub(/\$\{\{\s*(matrix\.[\w.-]+|matrix)\s*\}\}/, '{{\1}}')
```
jradtilbrook, line 142: I think all it needs is actually just
```suggestion
        var_name
```

mcncl: Funnily enough, that's what my local change is/was before I headed off.

app/Gemfile.lock
toote, line 1: This file should not be changed in this change

toote (changes requested):

toote (approved):

PR #64, opened 2023-10-28
steps context support

toote:
If a step has an `id` it is possible that a later job tries to use it so save the exit code to get a `step` status.

Unfortunately, step outputs are only available for `uses` which are being handled in a very ad-hoc way at this time with no outputs whatsoever


james2791 (approved):
🚢

jradtilbrook:
The conflict was just a snapshot so I regenerated them on this branch. @toote I'm not sure if you'll be around much but would be good to merge this. I can merge it in an hour or so if you don't get to it first

jradtilbrook (approved):

PR #69, opened 2023-10-30
Modularization of `uses` translation

toote:
Converts the huge `case` statement into individual functions in their own class.

Also corrects a bug with golang's translation that was not passing through the version correctly.

While at it, merged in a single function in its own file to the new class.


james2791 (approved):

app/spec/lib/bk/compat/__snapshots__/examples/github-actions/go-test.yml.snap
```diff
@@ -8,7 +8,7 @@ steps:
   - go test -v ./...
   plugins:
   - docker#v5.7.0:
-      image: golang:latest
+      image: golang:1.20
```
pzeballos, line 11: why do we need to stick to that image version?

toote: I forgot to add the bugfix to the description of the PR, sorry!

pzeballos (approved):

PR #74, opened 2023-11-07
plugin -> plugins and image whitespace corrected

pbuckley:
I copy/pasted this example into a pipeline in my demo org and got this feedback from the editor to get it running, seems like just a typo and some spacing wasn't caught


jradtilbrook (approved):
Good catch 👍
This was our manual mappings to work from so not the output of this migration tool. Still good to have correct though

PR #77, opened 2024-02-08
SUP-1812: /spec and /examples alignment for easier test addition/updates

james2791:
Currently, the example workflow/action configuration lives in `examples/<provider>` - which would be nicer placed inside their relative section in `/spec/lib/bk/compat/<provider>` directory.

In this PR, I also have made each provider have their own directory for their tests - GHA with snapshots/expression parser tests, Circle with snapshots only (at this point)


toote (approved):
great work!

PR #79, opened 2024-02-20
SUP-1787: GHA Docs Tabulation

james2791:
Transformed the GHA docs into a markdown tabulated format, with each main `key` of a workflow and all sub-key support listed (with touched up notes from the original)


toote (changes requested):
I love the new format, unfortunately I don't think that the emojis chosen to indicate the support are good enough.

To start with, they depend a lot on the emoji support of your machine. On Linux they are probably done through Google's Noto font so they look this way:
image. And when I test on OS X, I have a lot of difficulty differentiating between yellow and green due to my color-blindness.

Maybe use something that can be understood even when seen in black-and-white, like :heavy_check_mark:, :cold_sweat: and :no_entry:  that could never be confused with each other

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
+The Buildkite Migration tool's currently supported (🟢), partially supported (🟡) and unsupported (🔴) properties in translation of GitHub Action workflows to Buildkite pipelines are listed below.
 
 > [!NOTE]  
 > The Buildkite Migration tool does not currently support [GitHub secrets](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions) stored within GitHub organizations or repositories (such as `{{ secrets.FOO }}`). For securely storing secrets that can be used within a Buildkite pipeline - the [S3 Secrets Buildkite plugin](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks) can be setup/installed within a Buildkite agent (installed automatically within [Elastic CI Stack](https://github.com/buildkite/elastic-ci-stack-for-aws) setups). Additionally, Buildkite supports using a [secret storage service](https://buildkite.com/docs/pipelines/secrets#using-a-secrets-storage-service) through the use of dedicated plugins such as [HashiCorp Vault](https://github.com/buildkite-plugins/vault-secrets-buildkite-plugin) and [AWS Systems Manager Parameter Store](https://github.com/buildkite-plugins/aws-ssm-buildkite-plugin).
```
toote, line 6: I know that you have not changed this notice, but maybe change its wording to the fact that secret handling in BK is different and, instead of being part of the platform itself, for security reasons relies on plugins and external providers

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
-
-While there isn't a direct transition from a GitHub Action workflow to a Buildkite pipeline, Buildkite pipeline definitions allow for the utilisation of [YAML anchors](https://buildkite.com/docs/plugins/using#using-yaml-anchors-with-plugins) in which an anchor can be reused throughout various step definitions within a pipeline.
+## Concurrency (`concurrency`)
 
-Alternatively, the Buildkite platform allows jobs to have default configuration applied to them through the use of [Agent](https://buildkite.com/docs/agent/v3/hooks#agent-lifecycle-hooks) or [Job](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) Lifecycle hooks. Agent lifecycle hooks allows the configuration of customized commands on agent start up and shut down. Job Lifecycle hooks allow for the customization of what occurs at each stage of a job's lifecycle - and can be specified and applied at an agent, repository or plugin level.
+| Key | Supported? | Notes |
+| --- | ---------- | ----- |
+| `concurrency` | 🔴 | The Buildkite Migration tool supports the transition of the `concurrency` key (and `group`/`cancel-in-progress`) keys in job definitions that make up a workflow. |
```
toote, line 12: The icon says no, but the text says yes :thinking:

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
 
-### jobs
+## Defaults (`defaults`)
 
-GitHub Action workflows allows you to specify one or more `jobs` - main tasks of a workflow that run in parallel by default.
+| Key | Supported? | Notes |
+| --- | ---------- | ----- |
+| `defaults.run` | 🔴 | Buildkite pipeline definitions allow for common pipeline configuration to be applied with [YAML anchors](https://buildkite.com/docs/plugins/using#using-yaml-anchors-with-plugins), as well as setting up customised [agent](https://buildkite.com/docs/agent/v3/hooks#agent-lifecycle-hooks) and [job](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) lifecycle hooks. |
```
toote, line 18: But we do actually have _some_ `default` support: https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/parsers/gha.rb#L44

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
\ No newline at end of file
+> [!NOTE]  
+> When Buildkite builds are ran; each created command step inside the pipeline are ran as a [job](https://buildkite.com/docs/pipelines/defining-steps#job-states) that will be distributed and assigned to the matching agents meeting its specific queue and/or tag [targeting](h.ttps://buildkite.com/docs/pipelines/defining-steps#targeting-specific-agents). Each job is run within its own seperate environment, with potentially different environment variables (for example those defined at [step](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) level) - and is not always gauarnteed to run on the same agent depending on targeting rules specified/agent fleet setup.
```
toote, line 29: Minor wording change
```suggestion
> When Buildkite builds are ran; each created command step inside the pipeline is ran as a [job](https://buildkite.com/docs/pipelines/defining-steps#job-states) that will be distributed and assigned to the matching agents meeting its specific queue and/or tag [targeting](h.ttps://buildkite.com/docs/pipelines/defining-steps#targeting-specific-agents). Each job is run within its own seperate environment, with potentially different environment variables (for example those defined at [step](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) level) - and is not always gauarnteed to run on the same agent depending on targeting rules specified/agent fleet setup.
```

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
+
+| Key | Supported? | Notes |
+| --- | ---------- | ----- |
+| `jobs.<id>.concurrency` | 🟡 | The `group` name inside a `concurrency` definition inside a job maps to the `concurrency_group` [key](https://buildkite.com/docs/pipelines/controlling-concurrency#concurrency-groups) available within Buildkite. The `cancel-in-progress` optional value maps to the Buildkite pipeline setting of [Cancel Intermediate Builds](https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds)  Buildkite also allows a upper limit on how much jobs are created through a single step  definition with the `concurrency` key: which is set as `1` by default (there isn't a translatable key within a GitHub Action workflow). |
```
toote, line 33: There is no period or separation of the last section, but I believe that there should be 3 distinct paragraphs altogether as they show something translated, something that requires manual intervention and something new that people may be interested in.

```suggestion
| `jobs.<id>.concurrency` | 🟡 | The `group` name inside a `concurrency` definition inside a job maps to the `concurrency_group` [key](https://buildkite.com/docs/pipelines/controlling-concurrency#concurrency-groups) available within Buildkite.

The `cancel-in-progress` optional value maps to the Buildkite pipeline setting of [Cancel Intermediate Builds](https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds) that you will need to activate in the pipeline settings.
Buildkite also allows a upper limit on how much jobs are created through a single step  definition with the `concurrency` key: which is set as `1` by default (there isn't a translatable key within a GitHub Action workflow). |
```

toote (changes requested):
Some comments on the content itself.

In addition to that, there should be some comment on support for translating GHA expressions. Basically, they are translated but not all variables are available.

docs/GHA.md
james2791: Adjusted 👍

docs/GHA.md
james2791: Had to do this via the line break `<br/>` way but added here, and to the topmost note

docs/GHA.md
james2791: Tweaked the wording - technically we dont @ top level (as GHA allows CGs to be set to all jobs via this key, BK is in the specific command step which we parse 👍)

docs/GHA.md
james2791: Adjusted; primarily focused on secrets deliberately being decoupled from the core platform - and secrets service/plugin usage needed!

james2791:
@toote I've tweaked the docs to use a combination of ✅, ⚠️ and ❌ to compensate for the coloured circles for now - I think they work better - though let me know your opinion. Initially I had the uncoloured cross/check though these at least give a visual indicator with shape, but also if someone is viewing it with Noto etc)

docs/GHA.md
toote: I see what you are trying to say here.

docs/GHA.md
```diff
@@ -1,60 +1,66 @@
...
+## Concurrency (`concurrency`)
 
-Alternatively, the Buildkite platform allows jobs to have default configuration applied to them through the use of [Agent](https://buildkite.com/docs/agent/v3/hooks#agent-lifecycle-hooks) or [Job](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks) Lifecycle hooks. Agent lifecycle hooks allows the configuration of customized commands on agent start up and shut down. Job Lifecycle hooks allow for the customization of what occurs at each stage of a job's lifecycle - and can be specified and applied at an agent, repository or plugin level.
+| Key | Supported? | Notes |
+| --- | ---------- | ----- |
+| `concurrency` | ❌ | The Buildkite Migration tool supports the transition of the `concurrency` key (and `group`/`cancel-in-progress`) keys in `job` definitions that are mapped to specific [command steps](https://buildkite.com/docs/pipelines/command-step) of a [Buildkite pipeline](https://buildkite.com/docs/pipelines/controlling-concurrency#concurrency-groups). This isn't applicable to the entire pipeline and its steps which this key respects within GitHub Action workflows specifically. |
```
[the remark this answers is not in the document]
james2791: Tweaked 👍

docs/GHA.md
james2791: For now I'll 'resolve' this given we're going to look at `defaults` separately; added this PR as the reference to look into 👍

PR #82, opened 2024-02-28
SUP-1860: Some more GHA examples

james2791:
Some more GHA example added to spec, these being more of cases that include top-level concurrency, `defaults`, `max-parallel` (matrix) and some more app specific examples (XCode/RoR)


app/spec/lib/bk/compat/gha/examples/ruby-rails-full.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Create DB
+        run: |
+          bin/rails db:prepare
+
+      - name: Run tests
+        env:
+          KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: ${{ secrets.KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC }}
+          KNAPSACK_PRO_CI_NODE_TOTAL: ${{ matrix.ci_node_total }}
+          KNAPSACK_PRO_CI_NODE_INDEX: ${{ matrix.ci_node_index }}
```
toote, line 54: this example does not have a matrix configuration so I'm not sure what these variables are supposed to do here

toote (approved):
great examples, just a minor comment on a matrix-less job that uses matrix variables

app/spec/lib/bk/compat/gha/examples/ruby-rails-full.yaml
james2791: Did some tweaks to this - essentially the reference of Knapsack node references in CI, though pretty sure I tweaked this from a matrix reference. Removed them 👍

PR #85, opened 2024-03-04
SUP-1885: `working_directory` for CircleCI jobs and command ordering placement

james2791:
Fixed the ordering of parsing working directory (`cd`) step generation and merging.

Essentially we parse the generated steps through `LIST_ATTRIBUTES` - which includes the `commands` key. Upon generation of this step via `parse_executor`, we [merge](https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/parsers/circleci/jobs.rb#L29-L34) (`<<`) this created step the `bk_step` - the latter of which is generated first by `parse_command`.

With the merge, the commands are added at the _end_ of the commands list - this corrects it by determining if we've generated a step with a commands list of size 1, and starts with `cd`- for which is unshifted to the start of the list. Else - we concat the steps as per normal.

### DIFF
--- a/app/lib/bk/compat/pipeline/step.rb
+++ b/app/lib/bk/compat/pipeline/step.rb
@@ -141,7 +141,15 @@ module BK
       end
 
       def merge!(new_step)
-        LIST_ATTRIBUTES.each { |a| send(a).concat(new_step.send(a)) }
+ #       puts "new step"
+ #       p new_step.commands
+        LIST_ATTRIBUTES.each { |a| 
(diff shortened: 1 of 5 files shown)


james2791:
Il making this fix more portable just than inserting a check to the `merge` def at the moment

toote:
I created #86 to solve this same issue

james2791:
Merged #86 as above, so this can be closed - that adjusts the pre merge / `>>`

PR #86, opened 2024-03-06
SUP-1885: `working_dir` command ordering in CircleCI

toote:
Same as #85, but instead of having special treatment for step's command I found the bug that made the usage of `>>` not work and fixed it.

This way the fix of the issue is a single line... at the expense of having to refactor the `>>` method and everything that it touches :( On the bright side, we solved two bugs instead of just one :stuck_out_tongue:


james2791 (approved):
Nice 👍 Love how this handles the pre-merge!

PR #87, opened 2024-03-06
Remove unnecessary environment class

toote:
The Environment class provides a lot of complexity that is no longer being used and could be replaced with a straight Hash (proof of that is that no spec snapshots were changed :) )


james2791 (approved):
💚 it!

PR #90, opened 2024-03-19
[Improvements] Added .ruby-version for `rbenv` local use

james2791:
For folks using `rbenv`, you can set a local version specific to a repo/dir on the host to switch versions to/from

I'm of this type - but for anyone else running/managing the CLI with `rbenv` with a local version set in the migrations repo, this dotfile will be referred to when using the tool!


toote:
why not just add the `.ruby-version` file to the repository instead? That way people using rbenv will be able to take advantage of it

james2791:
> why not just add the `.ruby-version` file to the repository instead? That way people using rbenv will be able to take advantage of it

I've now accounted for the other way - _adding_ the dotfile! 😄

toote (commented):
I'd rather the version matches the one in the Dockerfile but definitely not critical at this time

toote (approved):
(sorry meant to approve it while leaving the comment)

james2791:
I've also matched the base image back to `3.2.2` - matching the dotfile add 👍

PR #91, opened 2024-03-20
Harness scaffolding

toote:
This creates the very basic scaffolding to start translating Harness CI files.

It is extremely bare-bones and the translation is awful and not correct, but should provide a base to start improving.

### DIFF
--- a/app/lib/bk/compat/parsers/harness.rb
+++ b/app/lib/bk/compat/parsers/harness.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+require_relative '../translator'
+require_relative '../pipeline'
+require_relative '../pipeline/step'
+require_relative './harness/steps/run'
+require_relative './harness/steps/background'
+
+module BK
+  module Compat
+    # Harness translation scaffolding
+    class Harness
+      include StepTranslator
+      require 'yaml'
+
+      def self.name
+        'Harness'
+      end
+
+      def self.option
+        'harness'
+      end
+
+      def self.matches?(text)
+        # sorting is important
+        config = YAML.safe_load(text, aliases: true)
+        mandatory_keys = %w[pipeline].freeze
+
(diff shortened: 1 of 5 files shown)


app/lib/bk/compat/parsers/harness.rb
```diff
@@ -0,0 +1,87 @@
...
+      def self.matches?(text)
+        # sorting is important
+        config = YAML.safe_load(text, aliases: true)
+        mandatory_keys = %w[pipeline].freeze
+
+        if config.is_a?(String)
+          false
+        elsif mandatory_keys & config.keys == mandatory_keys
+          config['pipeline'].include?('stages')
+        else
```
james2791, line 33: Rubocop not liking the `else` here in CI

app/lib/bk/compat/parsers/harness.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require_relative '../translator'
+require_relative '../pipeline'
+require_relative '../pipeline/step'
+require_relative './harness/steps/run'
+require_relative './harness/steps/background'
```
james2791, line 7: No `background.rb` in harness/steps dir - `spec` passes with this commented out/removed

james2791 (changes requested):

app/lib/bk/compat/parsers/harness.rb
toote: Made me think how to rewrite it to avoid the complaint, but I think I got it

app/lib/bk/compat/parsers/harness.rb
toote: :facepalm: this was left from some tests while doing some debugging (because obviously it took me quite a while to get this working)

james2791 (approved):
:shipit:

PR #93, opened 2024-04-22
BitBucket bootstrap

toote:
Creates the scaffolding for BitBucket translations.

Right now only supporting the basic running step with no additional attributes (just the mandatory script)


james2791 (approved):
:shipit:

PR #94, opened 2024-04-23
Image (`image_name`) parameter support (step/pipeline level)

james2791:
The `image` property can be set at both `step` and `pipeline` (root) level - the former proceeding the latter.

This PR adds support for either specifying the `image` at both these levels (can be set only at pipeline level, only at step level, or a combination of the two). Tests added to cater for each scenario. 

Note this is the `image_name` property as defined in [schema](https://bitbucket.org/atlassianlabs/intellij-bitbucket-references-plugin/raw/master/src/main/resources/schemas/bitbucket-pipelines.schema.json) - other types will need to be additive (AWS, with auth etc)


james2791:
Ill rebase this once [93](https://github.com/buildkite/migration/pull/93) is in

app/spec/lib/bk/compat/bitbucket/examples/image_pipeline_definition.yml
toote, line 1: I believe that this file is not needed, the `image_pipeline_and_step_definition.yml` one takes care of the situation as well

james2791: Removing this in favour for the _both_ step/pipeline test case 👍

app/spec/lib/bk/compat/bitbucket/examples/image_step_definition.yml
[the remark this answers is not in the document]
james2791: Removing this in favour for the _both_ step/pipeline test case 👍

PR #97, opened 2024-04-23
Pipes (un)supported

toote:
Scripts in steps can be a call to _pipes_ (atlassian's version of plugins or actions).

Added translation of them to be unsupported... for now


james2791 (approved):
✅

PR #99, opened 2024-04-23
SUP-2097: `after-script` translation

james2791:
Translating the `after-script` property of `step`s: which are equivalent to pre-exit hooks; and specifically repository located hooks.

### DIFF
--- a/app/lib/bk/compat/parsers/bitbucket/steps.rb
+++ b/app/lib/bk/compat/parsers/bitbucket/steps.rb
@@ -29,7 +29,7 @@ module BK
             agents: translate_agents(step.slice('size', 'runs-on'))
           ).tap do |cmd|
             cmd.timeout_in_minutes = step.fetch('max-time', nil)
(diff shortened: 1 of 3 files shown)


app/lib/bk/compat/parsers/bitbucket/steps.rb
```diff
@@ -29,6 +29,7 @@ def translator(conf, *, **)
             agents: translate_agents(step.slice('size', 'runs-on'))
           ).tap do |cmd|
             cmd.timeout_in_minutes = step.fetch('max-time', nil)
+            translate_after_script(cmd) if !step['after-script'].nil?
```
toote, line 32: I would inline the `cmd.add_commands` call here as it is a single line

toote (changes requested):

app/lib/bk/compat/parsers/bitbucket/steps.rb
james2791: Switched 👍

toote (approved):

toote (approved):

PR #102, opened 2024-04-24
SUP-2109: `oidc` translation 

james2791:
`oidc` parameter translation. Translates to our `aws-assume-role-with-web-identity` plugin, though the IAM role that is needed by the plugin can be specified in various areas (i.e via a flag calling `aws sts assume-role-with-web-identity`, or via an `AWS_ROLE_ARN` exported variable in the `step`'s `script`


james2791:
Closing as this was covered in #109

PR #103, opened 2024-04-24
BitBucket clone options

toote:
Translating clone options suggesting usage of agent configurations (except when using sparse checkouts that gets easily translated to a plugin).


james2791 (approved):
✅

PR #106, opened 2024-04-25
BitBucket conditionals

toote:
Translate file-filtering for steps as a (n untested) short-circuit command at the beginning of the step.


james2791 (approved):
🚢

PR #107, opened 2024-04-25
Global clone option

toote:
Adds `clone` to the list of defaults to use and an example to ensure it is correctly merged down to individual steps


james2791 (approved):
:shipit: (target branch to be one of thats merged to `main` soon)

PR #109, opened 2024-04-25
OIDC support

toote:
Adds support for when `oidc` is enabled by getting the token off buildkite's OIDC integration through the agent and keeping the environment variable name to prevent further modifications to the commands.

Closes #102 because it implements the same thing


james2791 (approved):
:shipit:

PR #111, opened 2024-04-27
Global caches

toote:
Adds a rough implementation for bitbucket's custom caches.

Note that:
* multi-file cache keys are translated to multiple cache entries which is less than ideal
* no-file caches are translated to pipeline-level cache configurations


james2791 (approved):
:shipit:

PR #112, opened 2024-04-28
Services first approach

toote:
Not a perfect translation but a good first step of using docker compose for service definitions.


james2791 (approved):
:shipit:

PR #113, opened 2024-04-28
Custom pipelines

toote:
Adds parsing of variables which are only allowed in custom pipelines (both in `pipelines.custom` as well as `definitions.pipelines`).

Not a perfect implementation but a good first step.


james2791 (approved):
:shipit:

PR #114, opened 2024-04-28
Stages support

toote:
Added support for stages that are just a container of steps.

Had to do some refactoring to take advantage of existing code.


james2791 (approved):
:shipit:

PR #115, opened 2024-04-28
Parallel steps

toote:
Parallel support was really easy because it was what we were already doing :facepalm: 

Had to rework a few bits for regular steps instead, which lead to all snapshots needing some modifications :cry:


james2791 (approved):
🚢

PR #118, opened 2024-04-30
Translator registration defs per parser

james2791:
We've got quite a few translators matching specific methods that are used within each pipeline parser - namely that of `translate_step` thats shared between all parsers currently.

Split out each parser's translator registration into its own `register_translators` def (privated) which are loaded on each class's `initialize`/creation.

### DIFF
--- a/app/lib/bk/compat/parsers/bitbucket.rb
+++ b/app/lib/bk/compat/parsers/bitbucket.rb
@@ -42,8 +42,6 @@ module BK
       end
 
       def parse
-        p @config
-        s = @config.slice('image', 'clone')]
         defaults = @config.slice('image', 'clone').merge(@config.fetch('options', {}))
         conf = @config['pipelines']
 
@@ -59,8 +57,8 @@ module BK
       def register_translators
         BK::Compat::BitBucketSteps::Import.new(register: method(:register_translator))
         BK::Compat::BitBucketSteps::Parallel.new(register: method(:register_translator), recursor: method(:translate_step))
-        BK::Compat::BitBucketSteps::Stages.new(register: method(:register_translator),recursor: method(:translate_step))
-        BK::Compat::BitBucketSteps::Step.new(register: method(:register_translator),definitions: @config.fetch('definitions', {}))
+        BK::Compat::BitBucketSteps::Stages.new(register: method(:register_translator), recursor: method(:translate_step))
+        BK::Compat::BitBucketSteps::Step.new(register: method(:register_translator), definitions: @config.fetch('definitions', {}))
         BK::Compat::BitBucketSteps::Variables.new(register: method(:register_translator))
       end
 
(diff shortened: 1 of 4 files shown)


app/lib/bk/compat/parsers/bitbucket.rb
```diff
@@ -67,6 +54,14 @@ def parse
 
       private
 
+      def register_translators
```
toote, line 57: due to ruby's conventions, methods that modify an objects state should end with `!`:
```suggestion
      def register_translators!
```
Note that this should be applied to all parsers

app/lib/bk/compat/parsers/bitbucket.rb
```diff
@@ -67,6 +54,14 @@ def parse
...
+        BK::Compat::BitBucketSteps::Import.new(register: method(:register_translator))
+        BK::Compat::BitBucketSteps::Parallel.new(register: method(:register_translator), recursor: method(:translate_step))
+        BK::Compat::BitBucketSteps::Stages.new(register: method(:register_translator), recursor: method(:translate_step))
+        BK::Compat::BitBucketSteps::Step.new(register: method(:register_translator), definitions: @config.fetch('definitions', {}))
```
toote, line 61: I'd suggest keeping the original formatting to avoid lines being way too long:
```suggestion
        BK::Compat::BitBucketSteps::Parallel.new(
          register: method(:register_translator),
          recursor: method(:translate_step)
        )
        BK::Compat::BitBucketSteps::Stages.new(
          register: method(:register_translator),
          recursor: method(:translate_step)
        )
        BK::Compat::BitBucketSteps::Step.new(
          register: method(:register_translator),
          definitions: @config.fetch('definitions', {})
        )
```

app/lib/bk/compat/parsers/circleci.rb
```diff
@@ -65,6 +62,10 @@ def parse
 
       private
 
+      def register_translators
+        BK::Compat::CircleCISteps::Builtins.new(register: method(:register_translator), recursor: method(:translate_steps))
```
toote, line 66: same here:
```suggestion
        BK::Compat::CircleCISteps::Builtins.new(
          register: method(:register_translator),
          recursor: method(:translate_steps)
        )
```

toote (approved):
Just a few minor suggestions for conventions and style

app/lib/bk/compat/parsers/bitbucket.rb
james2791: Applied 👍

app/lib/bk/compat/parsers/bitbucket.rb
james2791: Have reverted to the multiline way (agreed that its easier on the eyes!)

app/lib/bk/compat/parsers/circleci.rb
james2791: Have reverted to the multiline way (agreed that its easier on the eyes!)

PR #120, opened 2024-05-02
Dependency and Ruby `3.3.1` uplift across CI, release pipelines & image

james2791:
`ruby:3.3.1` released last week which was previously a blocker to us for anonymous splat forwarding in various `defs`.

Updated both CI/release pipelines to use `3.3.1`, image uplifted and various dependencies via `bundle`r bumped


toote (changes requested):
2 comments:
1- You forgot to update `.ruby-version` for consistency across the whole repository
2- (I believe because of 1-) building the image generates a warning: `Bundler 2.5.9 is running, but your lockfile was generated with 2.4.10.`

other than that, it works fine (both CLI and web interface)

james2791:
After doing a `ruby-local` 🐰 🕳️ - matched `bundler` to what is available in `ruby:3.3.1` containers, and also aligned the local `.ruby-version` after updating on my side/conforming CI works 👍

toote (approved):
:clap:

PR #122, opened 2024-05-09
Translator refactoring

toote:
The translation Mixin approach made the registration quite cumbersome and convoluted so I changed it from a module to a class to be instantiated. This would also allow to have multiple translators available per class in the future if/when necessary.


james2791 (approved):
:shipit: 

Definitely makes the translator impl/reading much nicer!

PR #123, opened 2024-05-09
Harness approval step

toote:
Implements Harness approval steps. Had to tweak a bit the logic of building the pipeline which led to simplified structure in the end :)

Refactored Harness translation to multiple files using a similar pattern as GHA does.


james2791 (approved):
💚

PR #124, opened 2024-05-10
Steps classes refactoring

toote:
Creates a better class hierarchy for steps to factor out common functionality.

Had to make the following changes just to reduce the changes to snapshots that I am not 100% happy about:
* custom key order for `CommandStep`
* change `<<` and `>>` to use `if` statements because `case/when` does not support inheritance :(


james2791 (approved):
Neat! :shipit:

PR #125, opened 2024-05-10
Stricter linter

toote:
I have long been running `rubocop` without `-l` as a stricter linter variant that mostly complains about:

* code complexity
* block length
* trailing whitespace

This PR includes refactorings to remove the last 8 warnings (and ignores one).

To further prevent this from being an issue I created an [EditorConfig](https://editorconfig.org/) file to prevent the easier automated warnings (as long as your IDE supports it, which it should - maybe with an extension)


james2791 (approved):
:shipit: 💚

PR #130, opened 2024-06-18
CircleCI when steps

toote:
CircleCI when steps in jobs can also call pre-defined commands, and that was not being processed correctly with the following error:

```
Failure/Error: commands = config['steps'].map { |s| recursor(*s.to_a.flatten) }
     
     NoMethodError:
       undefined method `to_a' for an instance of String
```

To solve this I needed the translation's `recursor` to be something else and not just a single step translation that did not have access to the context of the pipeline being parsed so I just made that customizable. After that, the code to handle the situation already existed in the job parsing function so I moved it to the general class file and used it to instantiate the translator.

Obviously, I updated the test for `when` conditionals with a minor change to also include this scenario.

### DIFF
--- a/app/lib/bk/compat/parsers/circleci.rb
+++ b/app/lib/bk/compat/parsers/circleci.rb
@@ -62,7 +62,7 @@ module BK
       private
 
       def register_translators!
-        @translator = BK::Compat::StepTranslator.new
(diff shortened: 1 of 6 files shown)


james2791 (approved):
👍 :shipit:

PR #131, opened 2024-06-20
SUP-2350 - Matrix and jobs via workflow docs note

james2791:
The migration tool will soon support matrix definitions when defined through a specific `job` as part of a `workflow`. Folks will need to map this in manually for now through to a [build matrix](https://buildkite.com/docs/pipelines/build-matrix) - which #132 adds to translation


james2791:
Accompanies #132

pzeballos (approved):

PR #135, opened 2024-06-24
SUP-2373: `image` translation (CI `Run` type step)

james2791:
CI Run [Steps](https://developer.harness.io/docs/continuous-integration/use-ci/run-step-settings) can have a specified `image` (+ the accompanying connector) to run said steps' command(s) within a container of said image.

Added image translation to `translate_run`, which I've also made a little bit modular with `post_keys` (eventually `pre-keys` too I imagine similar to how we do BBP step parsing).

Spec updated also to include command changes (adding Docker plugin)


app/lib/bk/compat/parsers/harness/steps/run.rb
```diff
@@ -8,10 +8,33 @@ module HarnessSteps
...
+        end
+
+        def translate_image(image)
+          return nil if image.nil?
+
+          BK::Compat::Plugin.new(
+            name: 'docker',
+            config: {
+              'image' => image
+            }.compact
```
toote, line 37: the `compact` here is not necessary as the code flow guarantees no values will be `nil` in this hash

```suggestion
            }
```

toote (approved):
just a minor comment here

app/lib/bk/compat/parsers/harness/steps/run.rb
james2791: 🙏

PR #136, opened 2024-06-24
SUP-2374: `shell` translation (CI `Run` type step)

james2791:
Translates `shell` property defined in CI Run steps to callout that configuration options for default shell (`Bash`, `Sh` etc) should be configured at agent level.


toote (approved):
I would just add a situation in the example file where the `shell` option is not specified to ensure that we do not break that scenario (I tested the code as-is and it is not broken)

PR #137, opened 2024-06-24
SUP-2381: CI stage `Background` type step translation

james2791:
`Background` type CI stage steps don't quite have a Buildkite translation - they map closest to an `agent-startup` hook for bringing up any dependant services that are required by all following `command` steps (for example).

Tweaked the existing behaviour of translating the entire `spec` Hash of a `Background` type CI stage step to a warning that it should be configured as such.


james2791:
I've left the `label` and `key` here in translation of these Background steps just to have the outputted `command` step easier to identify - especially if a pipeline has a few Background steps that need to be setup as agent startup hooks

toote (changes requested):
While I think that the translation is OK, I believe it would be better translated as a docker compose file instead of agent-startup hooks. What do you think?

james2791:
Maybe it could be an either/or @toote? Perhaps those used to putting common configuration in hooks, agent-level startup hooks probably will be familiar in setup, though I can see the appeal and approach a compose file/images would do also. Happy to add that to the cmd line 👍

james2791:
This will be closed in favour of when #163 is merged in via a composable service

PR #139, opened 2024-06-25
SUP-2293: Spec unification and file path refactor

james2791:
With the move of the examples for each parser to the `spec/lib/bk/compat/...` dirs - the created snapshots from `rspec/snapshot` would be generated with the relative directory of where the examples were placed: for example the GHA snapshots were constructed as `spec/lib/bk/compat/gha/__snapshots__/spec/lib/bk/compat/gha/examples`.

I've now moved all parser examples back to the `examples` directory inside `app/`, and configured the snapshots directory which is shared between the parsers. Computed snapshots live in the `spec/lib/bk/compat/snapshots/examples/<parser>` directories.


toote (approved):
Great job! I am not sure, though, about moving the examples outside of the `spec` folder. How about using `app/spec/examples` instead?

james2791:
Fair callout @toote 👍 

Wherever the examples are placed - the path starting from `spec/` will be appended to the snapshot dir (even if we specify the output snapshot dir within `spec_helper` as this PR does). 

Effectively if they're located somewhere in `spec/`, there will always be that replication of paths, in the case of the examples living in `spec/examples`, the snapshot dir will produce them within `spec/lib/bk/compat/snapshots/spec/examples/<tool>` - probably something we can come back towards, but this at least gives a bit of a cleaner structure!

PR #141, opened 2024-06-27
SUP-2396: Bitrise parser

james2791:
Implemented the first pass of Bitrise pipeline parsing - catering (for now) `script`, `change-workdir` and `git-clone` type jobs and X-ary number of `workflows`.

The only parameter that determines if a YAML pipeline is `format-version` - which is catered for upon users not specifying the parser in CLI options.

From the work done on steps above, a `BK::Compat::BitriseSteps::Translator` class for step translation is embedded that accounts for each of the types, with the specified `inputs` for each.


app/lib/bk/compat/parsers/bitrise/loader.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module BK
+  module Compat
+    # Bitrise workflow configuration loader
+    class Bitrise
+      STEP_KEY_REGEX = /([a-zA-Z-]*)@(\d*|0)(\.((\d*)|0)){0,3}/
```
toote, line 7: why the `\d*|0`? wouldn't it be easier to use `\d+`?

```suggestion
      STEP_KEY_REGEX = /([a-zA-Z-]*)@(v?\d+(\.\d+){0,3})?/
```

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -0,0 +1,16 @@
...
+module BK
+  module Compat
+    # Bitrise translation of workflows
+    class Bitrise
+      private
+
+      def parse_workflow(wf_name, _wf_config)
+        raise "No key within loaded workflow configuration for #{wf_name}" unless @steps_by_key.include?(wf_name)
+
+        @steps_by_key[wf_name]
```
toote, line 12: is there a reason to have this indirection? why not just parse the workflows directly into the step instead of loading and then just referencing them?

app/lib/bk/compat/parsers/bitrise/loader.rb
```diff
@@ -0,0 +1,39 @@
...
+      def load_workflow_step(step_key, step_inputs)
+        step_key_normalized = normalize_step_key(step_key)
+
+        # Translate step with normalized key (for translator) and its input configuration
+        @translator.translate_step(step_key_normalized, step_inputs)
+      end
+
+      def normalize_step_key(step_key)
+        # Pull out the step key without its version to match translator's supported steps
+        step_key.match?(STEP_KEY_REGEX) ? step_key.match(STEP_KEY_REGEX)[1] : step_key
```
toote, line 35: wouldn't it be easier to do
```suggestion
        step_key.split('@').first
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        def generate_change_workdir_command(path, is_create_path)
+          if is_create_path && path.present?
+            "mkdir #{path} && cd #{path}"
+          elsif !is_create_path && path.present?
+            "cd #{path}"
+          else
+            '# Invalid change-workdir step configuration!'
+          end
+        end
```
toote, line 34: being that `path` is mandatory I would refactor this two as:

```suggestion
        def translate_change_workdir(inputs)
          return '# Invalid change-workdir step configuration!' unless inputs.include?('path')
          
          [
            inputs['is_create_path'] ? "mkdir #{inputs['path']}" : nil,
            "cd #{inputs['path']}"
          ].compact
        end
```

app/lib/bk/compat/parsers/bitrise/loader.rb
```diff
@@ -0,0 +1,39 @@
...
+        @steps_by_key[key] = load_workflow_steps(key, config)
+      end
+
+      def load_workflow_steps(key, config)
+        # Extend for other types of steps - approvals/waits for example.
+        BK::Compat::CommandStep.new(label: key, key: key).tap do |cmd_step|
+          config['steps'].each do |step|
+            step_key, step_config = step.first
+            step_inputs = step_config['inputs'].reduce({}, :merge)
+            cmd_step.commands << load_workflow_step(step_key, step_inputs)
```
toote, line 21: the `.commands` is not necessary, the `CommandStep` class should do the right thing when you `<<` on it (taking care of flattening arrays and stuff like that):

```suggestion
            cmd_step << load_workflow_step(step_key, step_inputs)
```

app/lib/bk/compat/parsers/bitrise/steps.rb
toote: Note that this change would also avoid having to add `activesupport` as a dependency at all

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -0,0 +1,50 @@
...
+          [
+            ['# No need for cloning, the agent takes care of that']
+          ]
+        end
+
+        def translate_script(inputs)
+          [
+            inputs['content']
+          ]
+        end
```
toote, line 46: if the `.commands` is removed from the loader, you can remove the additional array:

```suggestion
        def translate_script(inputs)
          inputs['content']
        end
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -0,0 +1,50 @@
...
+          else
+            '# Invalid change-workdir step configuration!'
+          end
+        end
+
+        def translate_git_clone(_inputs)
+          [
+            ['# No need for cloning, the agent takes care of that']
+          ]
+        end
```
toote, line 40: if the .commands is removed from the loader, you can remove the additional arrays:

```suggestion
        def translate_git_clone(_inputs)
          '# No need for cloning, the agent takes care of that'
        end
```

toote (changes requested):
awesome bootstrapping of bitrise translation! A few suggestion to simplify the translation quite a bit (while also removing the need to add `activesupport` as a dependency)

app/lib/bk/compat/parsers/bitrise/loader.rb
james2791: Probably just a case of old cmd+c/v, as below though I think this was overthinking it anyway (unless we actually wanted to use the specific values of versions etc - as `String#split` works much better here anyway 👍

app/lib/bk/compat/parsers/bitrise/loader.rb
james2791: Works better than the regex/overthink - in latest 👍

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Altered - `<<` use w.r.t `CommandStep`s

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Altered - `<<` use w.r.t `CommandStep`s

app/lib/bk/compat/parsers/bitrise/loader.rb
james2791: Switched - the type from `load_workflow_step` will _now_ only be of `String` or `Array` class(es) - the latter for workdir step translation, but will apply for any others that need to pass in more than just one `command` for concetting

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Changed - and I also removed `activesupport` entirely in the latest commit (I'll still need to alter the behaviour in the other incoming parser steps in #147)

app/lib/bk/compat/parsers/bitrise/workflows.rb
james2791: The main thinking behind the `loader` is to effectively load in each `workflow`s' base configuration only (key point) to their specific steps (right now its just `CommandStep`s) - effectively its main responsibility as of now is loading the  commands by way of each step (and what is currently translatable)

It'll be the job of `parse_workflow` (and more future `def`s) that'll modify the base step through with additional properties (more example, dependencies etc) and tweak them before the parser renders out the configured pipeline. Effectively its 'similar' in a way to how Circle's parser does it through `@commands_by_key` - foundation layer (loader) and then adding bits/pieces to the config on-top

toote: I understand the similarities with CircleCI, but that was done that way because CircleCI's YAML has user-defined commands and jobs that can be instantiated in workflows (or other jobs/commands) so there was a clear dependency and reason for the loading. In this case, like any other parses that is not CircleCI, there is no such flexibility so I don't think that this is necessary

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -0,0 +1,41 @@
...
+
+        def translator(type, inputs)
+          send("translate_#{type.downcase.gsub('-', '_')}", inputs)
+        end
+
+        def translate_change_workdir(inputs)
+          generate_change_workdir_command(inputs)
+        end
+
+        def generate_change_workdir_command(inputs)
```
toote, line 22: why have a method that does nothing and just calls another one?

```suggestion
        def translate_change_workdir(inputs)
```

toote (changes requested):
just a minor suggestion, but I'm still not convinced of having a separate loader when it is not needed

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Looking at the code, I think that `generate_<step type>_command` might not actually work the best in the long run, it was more to split the underlying generation to defs to relay (and in the case of #151 thats split into a `commands.rb` just to make `steps.rb` not do SLOC ballooning)

I've reverted that decision in [cf59c3a](https://github.com/buildkite/migration/pull/141/commits/cf59c3a954c1381b4220f9b16976e0fb60c9efc0), will need to assess #151 based on the decision we take here

app/lib/bk/compat/parsers/bitrise/workflows.rb
james2791: Thinking more - I'm probably thinking two steps ahead from whats needed really - I reverted this back to the more simpler mechanism and not referencing an instance hash to pick out the steps per workflow.

For other more complex parsers - it might make sense, but its more simpler for now here 👍

toote (approved):
awesome work!

PR #147, opened 2024-07-03
SUP-2404: Bitrise `brew-install` step translation 

james2791:
[WIP]

(Will rebase this once #141 has gone in)


james2791:
With many additions and #141 soon going in, that shall be worked through first before this gets opened up

PR #151, opened 2024-07-08
SUP-2404: Bitrise `brew-instal`l (+ `bundler` and `git-tag`) step translation

james2791:
This PR introduces the translation of `brew-install` steps (along with pre-worked on `bundler` and `git-tag` step translations) - with a cleaner implementation from the simplicities that were merged with #141 (along with the main basis of #150)

All translations live in the translator (`steps.rb`) - and specs are refreshed with workflows embedded with `GroupStep`s


app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -7,6 +7,16 @@ class Bitrise
...
+        bk_steps = process_workflow_steps(wf_name, wf_config)
+
+        BK::Compat::GroupStep.new(
+          label: wf_name,
+          key: wf_name,
+          steps: [bk_steps]
+        )
+      end
+
+      def process_workflow_steps(wf_name, wf_config)
```
toote, line 19: this change means most examples end up with groups with single steps, which kind of defeat the purpose of the group. Or is there a reason for this that I'm missing?

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -28,6 +59,18 @@ def translate_git_clone(_inputs)
...
 
+        def translate_git_tag(inputs)
+          return '# Invalid git-tag step configuration!' unless inputs.include?('tag')
+
+          cmd_all_params = "git tag -fa #{inputs['tag']} -m #{inputs['tag_message']}"
+          cmd_no_message = "git tag -fa #{inputs['tag']}"
+
+          [
+            inputs['tag_message'].nil? ? cmd_no_message : cmd_all_params,
+            !inputs['push'].nil? && inputs['push'] ? 'git push --tags' : nil
```
toote, line 70: a simple way to write this would be:

```suggestion
            inputs.fetch('push', false) ? 'git push --tags' : nil
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -28,6 +59,18 @@ def translate_git_clone(_inputs)
...
         end
 
+        def translate_git_tag(inputs)
+          return '# Invalid git-tag step configuration!' unless inputs.include?('tag')
+
+          cmd_all_params = "git tag -fa #{inputs['tag']} -m #{inputs['tag_message']}"
+          cmd_no_message = "git tag -fa #{inputs['tag']}"
+
+          [
+            inputs['tag_message'].nil? ? cmd_no_message : cmd_all_params,
```
toote, line 69: in the [original code there are quotes around the tag message](https://github.com/Itelios/bitrise-steps-git-tag/blob/master/step.sh#L5) to ensure that, if not present it uses an empty string as the message and avoid the need for two different branches:

```suggestion
          [
            "git tag -fa #{inputs['tag']} -m \"#{inputs['tag_message']}\""
```

toote (changes requested):
some minor comments... but while we are at it, I may suggest adding some kind of reference to the original code of the steps for easier future reference.

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Oh yep one I missed 😮‍💨 - kept the initial ternary just on the case of no tag message; as the quotes would wrap a `nil`!

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: `Hash#fetch` is much nicer - I actually all similar conditionals to this approach!

app/lib/bk/compat/parsers/bitrise/workflows.rb
james2791: Good question @toote - I probably should have included this in the PR message - effectively this'll be the precursor to translating both `before_run` and `after_run` [params](https://devcenter.bitrise.io/en/references/workflows-reference.html) that a `workflow` can have defined alongside its `steps` (its about 6 examples down). Effectively, the two options run the `workflow` steps pre/post the `steps` of the actual workflow - thinking more to have these defined in the eventual translation as appending pre/post `CommandStep`s if we encounter either/or.

What do you think of this approach?

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -28,6 +59,18 @@ def translate_git_clone(_inputs)
...
 
+        def translate_git_tag(inputs)
+          return '# Invalid git-tag step configuration!' unless inputs.include?('tag')
+
+          cmd_all_params = "git tag -fa #{inputs['tag']} -m \"#{inputs['tag_message']}\""
+          cmd_no_message = "git tag -fa #{inputs['tag']}"
+
+          [
+            inputs.fetch('tag_message', false) ? cmd_all_params : cmd_no_message,
+            inputs.fetch('push', false) && inputs['push'] ? 'git push --tags' : nil
```
toote, line 70: the `inputs['push']` is now unnecessary:

```suggestion
            inputs.fetch('push', false) ? 'git push --tags' : nil
```

app/lib/bk/compat/parsers/bitrise/workflows.rb
toote: from what I've read, the `before_run` and `after_run` parameters define steps that need to run in the same execution environment as the rest of the workflow, which in BuildKite it translates to running inside the same step, so adding a GroupStep is not necessary as they would be commands within the same step anyways

app/lib/bk/compat/parsers/bitrise/steps.rb
toote: now that I think about it, if the default value is also false-y, then it should not be needed as `nil` would also map to the second branch of the conditional. As a general rule, `hash.fetch(key, default)` is only a good idea when `default` has a different behaviour than what would cause the key not being present

app/lib/bk/compat/parsers/bitrise/steps.rb
toote: that is weird,  I checked the behaviour in irb and it shouldn't be that way:

```
> inputs
=> {"tag"=>"v0.1.0"}
> "git tag -fa #{inputs['tag']} -m \"#{inputs['tag_message']}\""
=> "git tag -fa v0.1.0 -m \"\""
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -15,11 +15,42 @@ def translator(type, inputs)
...
+        def required_bundler_inputs(inputs)
+          inputs.include?('bundle_install_jobs') && inputs.include?('bundle_install_retry')
+        end
+
         def translate_change_workdir(inputs)
           return '# Invalid change-workdir step configuration!' unless inputs.include?('path')
 
           [
-            inputs['is_create_path'] ? "mkdir #{inputs['path']}" : nil,
+            inputs.fetch('is_create_path', false) ? "mkdir #{inputs['path']}" : nil,
```
toote, line 53: this `fetch` is not necessary as the default value maps to the same conditional branch as when the key is not present (and resolves to `nil`)

app/lib/bk/compat/parsers/bitrise/steps.rb
toote: and, as mentioned in the comments, if the default is the same as when the value is not present, the `fetch` is unnecessary:

```suggestion
            inputs['push'] ? 'git push --tags' : nil
```

toote (changes requested):

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: One I missed - cheers 👍

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Switched

app/lib/bk/compat/parsers/bitrise/workflows.rb
james2791: Thats fair - we'll assess this in due time but for now, I've wrapped everything back to singular `CommandStep`s in the latest commit!

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -15,11 +15,42 @@ def translator(type, inputs)
...
+        def required_bundler_inputs(inputs)
+          inputs.include?('bundle_install_jobs') && inputs.include?('bundle_install_retry')
+        end
+
         def translate_change_workdir(inputs)
           return '# Invalid change-workdir step configuration!' unless inputs.include?('path')
 
           [
-            inputs['is_create_path'] ? "mkdir #{inputs['path']}" : nil,
+            inputs.fetch('is_create_path') ? "mkdir #{inputs['path']}" : nil,
```
toote, line 53: this change is not necessary as `X.fetch(Y)` is the same as `X[Y]`

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -28,6 +59,18 @@ def translate_git_clone(_inputs)
...
         end
 
+        def translate_git_tag(inputs)
+          return '# Invalid git-tag step configuration!' unless inputs.include?('tag')
+
+          cmd_all_params = "git tag -fa #{inputs['tag']} -m \"#{inputs['tag_message']}\""
+          cmd_no_message = "git tag -fa #{inputs['tag']}"
+
+          [
+            inputs.fetch('tag_message', false) ? cmd_all_params : cmd_no_message,
```
toote, line 69: To be consistent with the original code, this should be simplified as:

```suggestion
          [
            "git tag -fa #{inputs['tag']} -m \"#{inputs['tag_message']}\"",
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -15,11 +15,42 @@ def translator(type, inputs)
...
+          return '# Invalid brew-install configuration!' unless inputs.include?('verbose_log')
+
+          generate_brew_install_command(inputs)
+        end
+
+        def generate_brew_install_command(inputs)
+          if inputs.fetch('packages', false)
+            install_type = inputs.fetch('upgrade', false) ? 'reinstall' : 'install'
+            "brew #{install_type}#{inputs['verbose_log'] ? ' -vd' : nil} #{inputs['packages']}"
+          elsif inputs.fetch('use_brewfile', false) && inputs.fetch('brewfile_path', false)
```
toote, line 28: This is the same as:

```suggestion
          if inputs['packages']
            install_type = inputs.fetch('upgrade', false) ? 'reinstall' : 'install'
            "brew #{install_type}#{inputs['verbose_log'] ? ' -vd' : nil} #{inputs['packages']}"
          elsif inputs['use_brewfile'] && inputs['brewfile_path']
```

toote (changes requested):
Just some minor corrections and we should be good to merge

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Adjusted back 👍

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Adjusted back 👍

app/lib/bk/compat/parsers/bitrise/steps.rb
james2791: Switched for consistency vs the ternary implementation

toote (approved):

PR #182, opened 2025-01-20
Sup 2923 install command

phanichowdary9988:
Add Build & install-docker commands


app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
           ]
         end
       end
+
+      # Handles Docker installation logic
+      class DockerInstaller
```
toote, line 24: while not against it, I don't think that this warrants a whole class on its own as it is only instantiated once to manipulate the config in a very straight-forward way

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+        DOCKER_INSTALL_PARAMS = {
+          'install-dir' => {
+            'type' => 'string',
+            'description' => 'Directory in which to install Docker binaries',
```
toote, line 28: what is this description for?

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+          },
+          'version' => {
+            'type' => 'string',
+            'description' => 'Version of Docker to install',
+            'default' => 'latest'
+          }
+        }.freeze
+
+        def initialize(config = {})
+          @config = config
```
toote, line 39: this class has a single purpose and its `config` is not an internal state

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+          [
+            generate_script_header,
+            generate_custom_dir_commands,
+            generate_version_specific_commands
+          ].flatten.compact
+        end
+
+        def generate_script_header
+          [
+            '#!/usr/bin/env bash',
```
toote, line 80: this project does not generate scripts so the shebang is definitely not necessary

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+        end
+
+        def generate_version_specific_commands
+          if specific_version?
+            generate_specific_version_install
+          else
+            [
+              'curl -fsSL https://get.docker.com -o get-docker.sh',
+              'sh get-docker.sh'
+            ]
```
toote, line 104: this code is duplicated from the `simple_install`

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+
+        def matcher(orb, _conf)
+          orb.start_with?('docker/')
+        end
+
+        def translator(action, config)
+          @config = config
+          method_map = {
+            'docker/build' => -> { translate_docker_build(config) },
+            'docker/install' => -> { translate_docker_install },
```
toote, line 136: the `docker/install` command does not exist in the original orb

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+
+        def translate_docker_build(config)
+          image_tag = "#{config['image']}:#{config['tag'] || 'latest'}"
+          [
+            '# No need to setup remote docker, use the host docker',
+            "docker build -t #{image_tag} ."
+          ]
+        end
+
+        def translate_docker_install
```
toote, line 152: this is the third time this piece of code exists in this file

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,144 @@ def translator(action, _config)
...
+          }
+          method_map.fetch(action, -> { ["# Unsupported docker orb action: #{action}"] }).call
+        end
+
+        private
+
+        def translate_docker_build(config)
+          image_tag = "#{config['image']}:#{config['tag'] || 'latest'}"
+          [
+            '# No need to setup remote docker, use the host docker',
```
toote, line 147: why is this comment here?

app/spec/lib/bk/compat/circleci_spec.rb
```diff
@@ -21,3 +21,143 @@
...
+                               "echo '~~~ Installing Docker'",
+                               'curl -fsSL https://get.docker.com -o get-docker.sh',
+                               'sh get-docker.sh'
+                             ])
+      end
+    end
+  end
+end
+
+RSpec.describe BK::Compat::CircleCISteps::DockerOrb do
```
toote, line 132: there is no need to test this class specifically, the snapshot testing takes care of that

app/spec/lib/bk/compat/circleci_spec.rb
```diff
@@ -21,3 +21,143 @@
     end
   end
 end
+
+RSpec.describe BK::Compat::CircleCISteps::DockerInstaller do
```
toote, line 25: if you remove the class, this should be removed as well. If not, I'd suggest moving it to its own spec file

toote (changes requested):
nice first approach.

In addition to the comments on particular lines/files, I'd also suggest:
* have the `install-docker` translation add a comment recommending that docker be installed and ready in agents instead of doing that in steps
* if you are supporting all options for the `install-docker` orb command, you should also attempt to support all the options of the build command (though, in this case it may be better to create a separate PR for that)

phanichowdary9988:
As suggested above fixed the comments and updated the PR

PR #184, opened 2025-01-23
Sup 2923 install docker

phanichowdary9988:
This PR is created to add install-docker command to support docker orb for Circleci

### DIFF
--- a/app/lib/bk/compat/parsers/circleci/orbs.rb
+++ b/app/lib/bk/compat/parsers/circleci/orbs.rb
@@ -27,7 +27,7 @@ module BK
         end
 
         def translator(action, config)
-          method_name = "translate_#{action.gsub('/', '_')}"
+          method_name = "translate_#{action.tr('/-', '_')}"
           return send(method_name, config) if respond_to?(method_name, true)
 
           unsupported_action(action)
--- a/app/spec/lib/bk/compat/snapshots/examples/circleci/install.yml.snap
+++ b/app/spec/lib/bk/compat/snapshots/examples/circleci/install.yml.snap
@@ -5,7 +5,11 @@ steps:
   - echo '~~~ Install Dependencies'
   - apt-get update
   - apt-get install -y curl jq git
-  - "# Unsupported docker orb action: docker/install-docker"
+  - "# Instead of installing docker in a step, we recommend your agent environment
+    to have it pre-installed"
+  - echo '~~~ Installing Docker'
+  - curl -fsSL https://get.docker.com -o get-docker.sh
+  - sh get-docker.sh
   plugins:
   - docker#v5.10.0:
       image: cimg/base:stable
@@ -17,7 +21,16 @@ steps:
   - echo '~~~ Install Dependencies'
   - apt-get update
   - apt-get install -y curl jq git
-  - "# Unsupported docker orb action: docker/install-docker"
+  - "# Instead of installing docker in a step, we recommend your agent environment
+    to have it pre-installed"
+  - echo '~~~ Installing Docker'
+  - mkdir -p /var/lib/docker
+  - VERSION=20.10.7
+  - curl -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${VERSION}.tgz
+    -o docker.tgz
+  - tar xzvf docker.tgz
+  - cp docker/* /var/lib/docker/
+  - rm -rf docker docker.tgz
   plugins:
   - docker#v5.10.0:
       image: cimg/base:stable


app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
...
+        def initialize
+          @config = {}
+        end
+
+        def matcher(orb, _conf)
+          orb.start_with?('docker/')
+        end
+
+        def translator(action, config)
+          @config = config
```
toote, line 34: if you are passing `config` through to the methods, the instance variable is not necessary

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
...
+            'docker/build' => -> { translate_docker_build(config) },
+            'docker/install-docker' => -> { validate_and_translate_install_docker(config) }
+          }
+          method_map.fetch(action, -> { ["# Unsupported docker orb action: #{action}"] }).call
+        end
+
+        private
+
+        def validate_and_translate_install_docker(_config)
+          install_dir = @config['install-dir'] || '/usr/local/bin'
```
toote, line 45: you should use the `config` parameter instead of the instance variable

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
...
+          version = @config['version'] || 'latest'
```
toote, line 46: just for consistency with the rest of the codebase I would recommend changing these constructs to use `fetch` instead (take into account a previous comment about not using the instance variable as well):

```suggestion
          install_dir = @config.fetch('install-dir', '/usr/local/bin')
          version = @config.fetch('version', 'latest')
```

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
...
+          commands = ['"# No need for install docker, the agent takes care of that"']
+          commands << "echo '~~~ Installing Docker'"
+          commands << "mkdir -p #{install_dir}" if install_dir != '/usr/local/bin'
+          commands += install_commands(version, install_dir)
+          commands
```
toote, line 51: You are creating a single array to return and operating on it 3 different ways. This same thing can be achieved with structures [similar to others used across this codebase](https://github.com/buildkite/migration/blob/main/app/lib/bk/compat/parsers/circleci/steps.rb#L68-L76):

```suggestion
          [
            '# No need for install docker, the agent takes care of that',
            "echo '~~~ Installing Docker'",
            install_dir != '/usr/local/bin' ? "mkdir -p #{install_dir}" : nil,
            install_commands(version, install_dir)
          ].flatten.compact
```

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
...
+        end
+
+        private
+
+        def validate_and_translate_install_docker(_config)
+          install_dir = @config['install-dir'] || '/usr/local/bin'
+          version = @config['version'] || 'latest'
+          commands = ['"# No need for install docker, the agent takes care of that"']
+          commands << "echo '~~~ Installing Docker'"
+          commands << "mkdir -p #{install_dir}" if install_dir != '/usr/local/bin'
```
toote, line 49: if `version` is `latest`, this directory is not used so it may make more sense to move this to the `install_commands` function

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,55 @@ def translator(action, _config)
           ]
         end
       end
+
+      # Docker orb specific implementation
+      class DockerOrb
+        def initialize
+          @config = {}
```
toote, line 26: this is not really necessary as the class doesn't need to keep any internal state

toote (changes requested):
excellent progress from #182, just some more comments that mostly have to do with code-styling and cleaning up the logic

phanichowdary9988:
> excellent progress from #182, just some more comments that mostly have to do with code-styling and cleaning up the logic

PR updated

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,53 @@ def translator(action, _config)
...
+        def matcher(orb, _conf)
+          orb.start_with?('docker/')
+        end
+
+        def translator(action, config)
+          method_map = {
+            'docker/build' => -> { translate_docker_build(config) },
+            'docker/install-docker' => -> { validate_and_translate_install_docker(config) }
+          }
+          method_map.fetch(action, -> { ["# Unsupported docker orb action: #{action}"] }).call
```
toote, line 34: very nice pattern but has the disadvantage of calling both translations just to create the mapping in addition to creating unnecessary pass-through lambdas. Instead I would do something like the following

```suggestion
          method_map = {
            'docker/build' => 'translate_docker_build',
            'docker/install-docker' => 'validate_and_translate_install_docker'
          }
          send(method_map.fetch(action, 'default_translation'), config)
```

and create a `default_translation` method in the class that returns the array.

If you rename the methods you can even do something like what is done [in the harness step translator](https://github.com/buildkite/migration/blob/819da4380308094ea7bb1caf31d80ac2b7ad8370/app/lib/bk/compat/parsers/harness/steps.rb#L19), [in the circleci executor translator](https://github.com/buildkite/migration/blob/819da4380308094ea7bb1caf31d80ac2b7ad8370/app/lib/bk/compat/parsers/circleci/executors.rb#L43) and [gha action translator](https://github.com/buildkite/migration/blob/819da4380308094ea7bb1caf31d80ac2b7ad8370/app/lib/bk/compat/parsers/gha/steps/actions.rb#L17)

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,53 @@ def translator(action, _config)
...
+        end
+
+        private
+
+        def validate_and_translate_install_docker(config)
+          install_dir = config.fetch('install-dir', '/usr/local/bin')
+          version = config.fetch('version', 'latest')
+
+          [
+            '# No need for install docker, the agent takes care of that',
```
toote, line 44: The agent doesn't actually take care of it

```suggestion
            '# Instead of installing docker in a step, we recommend your agent environment to have it pre-installed',
```

app/lib/bk/compat/parsers/circleci/orbs.rb
```diff
@@ -28,6 +19,53 @@ def translator(action, _config)
...
+            "echo '~~~ Installing Docker'",
+            install_dir == '/usr/local/bin' ? nil : "mkdir -p #{install_dir}",
+            install_commands(version, install_dir)
+          ].flatten.compact
+        end
+
+        def install_commands(version, install_dir)
+          (install_dir == '/usr/local/bin' ? [] : ["mkdir -p #{install_dir}"])
```
toote, line 52: this does nothing, does it?

toote (changes requested):
:clap: almost there!

app/spec/lib/bk/compat/snapshots/examples/circleci/install.yml.snap
```diff
@@ -20,15 +17,7 @@ steps:
...
-  - "# No need for install docker, the agent takes care of that"
-  - echo '~~~ Installing Docker'
-  - mkdir -p /var/lib/docker
-  - VERSION=20.10.7
-  - curl -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${VERSION}.tgz
-    -o docker.tgz
-  - tar xzvf docker.tgz
-  - cp docker/* /var/lib/docker/
-  - rm -rf docker docker.tgz
+  - "# Unsupported docker orb action: docker/install-docker"
```
toote, line 20: I thought the whole purpose of this PR was to support this action :P

toote (changes requested):

toote (approved):
:clap: thanks a lot for the effort to drive this through the finish line!

pzeballos (approved):

PR #188, opened 2025-01-31
SUP 2959 push-pull command

phanichowdary9988:
This PR is created to add push-pull commands to support docker orb for Circleci


app/lib/bk/compat/parsers/circleci/orbs/docker/pull_push.rb
```diff
@@ -0,0 +1,50 @@
...
+        def fetch_image(config)
+          config.fetch('image') { raise 'Image name must be provided' }
+        end
+
+        def fetch_tags(config)
+          return [ENV['CIRCLE_SHA1']] if config['tag'].nil? && ENV['CIRCLE_SHA1']
+
+          config.fetch('tag', '').split(',').map(&:strip).tap do |tags|
+            raise 'Tag or CIRCLE_SHA1 must be provided' if tags.empty?
+          end
```
toote, line 45: according to the [orb's documentation](https://circleci.com/developer/orbs/orb/circleci/docker#commands-push) the `tag` parameter is not required so this piece of code raises an exception for a valid configuration

toote (changes requested):
In addition to the specific comments, the files in the `orbs/docker` folder do not need to define their own classes, they can re-open the `DockerOrb` class and add methods to it, vastly simplifying all your code

app/lib/bk/compat/parsers/circleci/orbs/docker.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require_relative 'docker/docker'
+
+module BK
+  module Compat
+    module CircleCISteps
+      DockerOrb = BK::Compat::Parsers::CircleCI::Orbs::Docker::Docker
```
toote, line 8: the way it was before the last commit was OK, why did you remove the class only to add it as a reference here?

app/lib/bk/compat/parsers/circleci/orbs/docker/docker.rb
toote, line 1: This should have been the `app/lib/bk/compat/parsers/circleci/orbs/docker.rb` file

app/lib/bk/compat/parsers/circleci/orbs/docker/docker.rb
```diff
@@ -0,0 +1,51 @@
...
+require_relative 'commands/pull_push'
+
+module BK
+  module Compat
+    class Parsers
+      module CircleCI
+        module Orbs
+          module Docker
+            # Main Docker orb implementation for CircleCI translation
+            class Docker
```
toote, line 14: This should be a lot simpler, just like the when the class was in the same file as the `GenericOrb` class:

```suggestion
module BK
  module Compat
    module CircleCISteps
      # Main Docker orb implementation for CircleCI translation
      class DockerOrb
```

app/lib/bk/compat/parsers/circleci.rb
```diff
@@ -6,6 +6,7 @@
 require_relative 'circleci/executors'
 require_relative 'circleci/jobs'
 require_relative 'circleci/orbs'
+require_relative 'circleci/orbs/docker'
```
toote, line 9: this should not be necessary here, it should be included in the `circleci/orbs.rb` file instead

app/lib/bk/compat/parsers/circleci/orbs/docker/commands/pull_push.rb
```diff
@@ -0,0 +1,40 @@
...
+
+module BK
+  module Compat
+    class Parsers
+      module CircleCI
+        module Orbs
+          module Docker
+            # Handles docker pull and push commands for CircleCI orb
+            module Commands
+              module_function
```
toote, line 11: the fact that you had to use this means that you are not implementing a class here.

app/lib/bk/compat/parsers/circleci/orbs/docker/commands/pull_push.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module BK
+  module Compat
+    class Parsers
+      module CircleCI
+        module Orbs
+          module Docker
+            # Handles docker pull and push commands for CircleCI orb
+            module Commands
```
toote, line 10: You should be extending the `DockerOrb` class here:

```suggestion
module BK
  module Compat
    module CircleCISteps
      module BK
      # Docker orb pull and push commands
      class DockerOrb
```

toote (changes requested):

app/lib/bk/compat/parsers/circleci/orbs/docker.rb
```diff
@@ -0,0 +1,27 @@
...
+    module CircleCISteps
+      # Docker orb specific implementation
+      class DockerOrb
+        def matcher(orb, _conf)
+          orb.start_with?('docker/')
+        end
+
+        def translator(action, config)
+          method_name = "translate_#{action.tr('/-', '_')}"
+          respond_to?(method_name, true) ? send(method_name, config) : unsupported_action(action)
```
toote, line 18: the `true` in the `respond_to` is not necessary if the `translate_*` methods are not private:
```suggestion
          respond_to?(method_name) ? send(method_name, config) : unsupported_action(action)
```

app/lib/bk/compat/parsers/circleci/orbs/docker/build.rb
```diff
@@ -0,0 +1,58 @@
...
+            *generate_optional_args(config),
+            config.fetch('docker-context', '.')
+          ].compact.join(' ')
+        end
+
+        def generate_optional_args(config)
+          [
+            ("--cache-from=#{config['cache_from']}" if config['cache_from']),
+            ("--cache-to=#{config['cache_to']}" if config['cache_to']),
+            ('--load' if config['cache_to']),
```
toote, line 45: why is this option necessary?

app/lib/bk/compat/parsers/circleci/orbs/docker/build.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        def generate_build_command(config)
+          [
+            'docker',
+            ('buildx' if config['cache_to']),
+            'build',
+            '-f', File.join(config.fetch('path', '.'), config.fetch('dockerfile', 'Dockerfile')),
+            '-t', generate_tag(config),
+            ("--build-context source=#{config['attach-at']}" if config['attach-at']),
```
toote, line 35: taking into account the `attach-at` option appears to be specific to CircleCI (documentation states that it will mount a workspace previously saved with `persist_to_workspace`, which is not supported) I'm not sure that it is a good idea to translate it this way, but I'm not 100% against it either

app/lib/bk/compat/parsers/circleci/orbs/docker/pull_push.rb
```diff
@@ -0,0 +1,32 @@
...
+          images = config.fetch('images', '').split(',').map(&:strip)
+          return ['# No images provided to pull.'] if images.empty?
+
+          ignore_error = config['ignore-docker-pull-error'] ? ' || true' : ''
+          images.map { |image| "docker pull #{image}#{ignore_error}" }
+        end
+
+        def translate_docker_push(config)
+          full_image = config.fetch('image') { raise 'Image name must be provided' }
+          full_image = "#{config['registry']}/#{full_image}" if config['registry']
```
toote, line 18: I don't think it is a good idea to assign a variable only to reassign it immediately afterwards, a better way would be something like:

```suggestion
          raise 'Image name must be provided' unless config.include?('image')
          
          full_image = config.values_at('registry', 'image').compact.join('/')
```

toote (approved):
:+1: just a few minor suggestions or comments

app/lib/bk/compat/parsers/circleci/orbs/docker/pull_push.rb
phanichowdary9988: This change has raised a Abc size error when run rubocop. So, in addition to this change, slightly modified the translate_docker_push method.

PR #191, opened 2025-02-06
SUP 2961 & SUP 2964 install-docker-compose & install-dockerize commands

phanichowdary9988:
This PR is created to add install-docker-compose & install-dockerize  commands to support docker orb for Circleci


app/lib/bk/compat/parsers/circleci/orbs/docker/install_tools.rb
```diff
@@ -34,6 +58,30 @@ def install_commands(version, install_dir)
             ]
           end
         end
+
+        def install_docker_compose_commands(version, install_dir)
+          binary_url = 'https://github.com/docker/compose/releases/' \
+                       "#{version == 'latest' ? 'latest/' : "download/#{version}/"}docker-compose-linux-x86_64"
```
toote, line 64: what about OSs that are not linux and/or archs that are not x86_64?

toote (changes requested):
My only problem with the implementation is that it would only work in linux running on x86 64-bit architectures. If that is not something that can be solved, I'd rather the tool only printed the comment about it not being supported

app/lib/bk/compat/parsers/circleci/orbs/docker/install_tools.rb
```diff
@@ -34,6 +58,46 @@ def install_commands(version, install_dir)
             ]
           end
         end
+
+        def install_docker_compose_commands(version, install_dir)
+          platform = `uname -s`.strip.downcase
+          sys_arch = `uname -m`.strip.downcase
```
toote, line 64: the backticks would run when the conversion tool is run in - what one could assume would be - a dev's computer and not the same place where the steps will run so this would mean that there is a very likely architecture and OS mismatch

toote (changes requested):

app/lib/bk/compat/parsers/circleci/orbs/docker/install_dockerize.rb
```diff
@@ -0,0 +1,54 @@
...
+           '',
+           '  if [ -f /etc/issue ] && grep -q "Alpine" /etc/issue; then',
+           '    PLATFORM="alpine-linux-amd64"',
+           '  else',
+           '    PLATFORM="linux-${SYS_ARCH}"',
+           '  fi',
+           'fi',
+           '',
+           'BINARY_URL=""',
+           'if [ "$version" = "latest" ]; then',
```
toote, line 40: the `version` variable does not appear to be initialized to anything in this script

app/lib/bk/compat/parsers/circleci/orbs/docker/install_dockerize.rb
```diff
@@ -0,0 +1,54 @@
...
+
+        private
+
+        def install_dockerize_commands(_version, install_dir)
+          ['PLATFORM=""',
+           'if [ "$(uname -s)" = "Darwin" ]; then',
+           '  PLATFORM="darwin-amd64"',
+           'else',
+           '  SYS_ARCH=$(uname -m)',
+           '  case "$SYS_ARCH" in',
```
toote, line 27: in the docker compose script this is done with a regex, is there any reason for it to be different?

app/lib/bk/compat/parsers/circleci/orbs/docker/install_docker.rb
toote, line 1: most of the whitespace changes to this file look unnecessary (the renaming of the functions may be convenient to avoid clashing)

app/examples/circleci/install-docker-compos.yml
toote, line 1: not sure if it was on purpose but this should be named `install-docker-compose.yml`

app/lib/bk/compat/parsers/circleci/orbs/docker/install_docker_compose.rb
```diff
@@ -0,0 +1,40 @@
...
+          ].flatten.compact
+        end
+
+        private
+
+        def install_docker_compose_commands(_version, install_dir)
+          [
+            "PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')",
+            "SYS_ARCH=$(uname -m | tr '[:upper:]' '[:lower:]')",
+            'SYS_ARCH="${SYS_ARCH/arm64/aarch64}"',
```
toote, line 26: in the `dockerize` file you are doing the exact same thing with a `case` statement, why the difference?

app/lib/bk/compat/parsers/circleci/orbs/docker/install_docker_compose.rb
```diff
@@ -0,0 +1,40 @@
...
+            'if [ "$version" = "latest" ]; then',
```
toote, line 27: same as the other script, the `version` variable does not appear to be defined in this script

app/lib/bk/compat/parsers/circleci/orbs/docker/install_docker_compose.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Compat
+    module CircleCISteps
+      # Handles the translation of docker installation commands
+      class DockerOrb
+        def translate_docker_install_docker_compose(config)
+          install_dir = config.fetch('install-dir', '/usr/local/bin')
+          version = config.fetch('version', 'latest')
+          [
+            '# Instead of installing docker-compose in a step, ' \
+            'we recommend your agent environment to have it pre-installed',
```
toote, line 13: why have this particular message split into two lines whereas all the others are in a single line?

Either have all of them be in a single line or split them in the resulting script as well:

```suggestion
            '# Instead of installing docker-compose in a step, ',
            '# we recommend your agent environment to have it pre-installed',
```

toote (changes requested):
there are some problems with the scripts to install the tools and some wildly inconsistencies in the code itself that would be best to unify

app/lib/bk/compat/parsers/circleci/orbs/docker/install_docker_compose.rb
phanichowdary9988: For dockerize, packages are of the format, dockerize-platform-<amd64/arm64>. So, I need to map CPU architectures to match package names accordingly . (https://github.com/jwilder/dockerize/releases)
For docker-compose, all package names are already mapped to architectures except for arm64, So I have used single variable replacement.(https://github.com/docker/compose/releases/)

toote (approved):

PR #201, opened 2025-02-25
SUP_2406_github_status_translation

phanichowdary9988:
This PR is created to translate github_status for bitrise

### DIFF
--- a/app/lib/bk/compat/parsers/bitrise/github_status.rb
+++ b/app/lib/bk/compat/parsers/bitrise/github_status.rb
@@ -6,10 +6,13 @@ module BK
       # Handles the translation of github-status
       class Translator
         def translate_github_status(inputs)
-          raise '# Invalid github-status step configuration!' unless required_github_status_inputs(inputs)
+          raise 'Invalid github-status step configuration!' unless required_github_status_inputs(inputs)
 
           params = extract_github_status_params(inputs)
-          BK::Compat::CommandStep.new(notify: generate_github_status_notify(params))
+          {
+            'notify' => generate_github_status_notify(params),
+            'commands' => generate_github_status_command(params)
+          }
         end
 
         private
@@ -31,12 +34,18 @@ module BK
           [
             {
               'github_commit_status' => {
-                'context' => params[:context],
-                'description' => params[:description],
-                'status' => params[:status],
-                'target_url' => params[:target_url]
+                'context' => params[:context]
               }
             }
+          ]
+        end
+
+        # Command step to update commit status correctly
+        def generate_github_status_command(params)
+          [
+            "buildkite-agent meta-data set github_commit_status_description '#{params[:description]}'",
+            "buildkite-agent meta-data set github_commit_status_status '#{params[:status]}'",
+            "buildkite-agent meta-data set github_commit_status_target_url '#{params[:target_url]}'"
           ]
         end
 
--- a/app/lib/bk/compat/parsers/bitrise/github_status.rb
+++ b/app/lib/bk/compat/parsers/bitrise/github_status.rb
@@ -9,7 +9,7 @@ module BK
           raise '# Invalid github-status step configuration!' unless required_github_status_inputs(inputs)
 
           params = extract_github_status_params(inputs)
-          generate_github_status_notify(params)
+          BK::Compat::CommandStep.new(notify: generate_github_status_notify(params))
         end
 
         private
@@ -28,14 +28,16 @@ module BK
         end
 
         def generate_github_status_notify(params)
-          <<~YAML
-            notify:
-              - github_commit_status:
-                  context: "#{params[:context]}"
-                  description: "#{params[:description]}"
-                  status: "#{params[:status]}"
-                  target_url: "#{params[:target_url]}"
-          YAML
+          [
+            {
+              'github_commit_status' => {
+                'context' => params[:context],
+                'description' => params[:description],
+                'status' => params[:status],
+                'target_url' => params[:target_url]
+              }
+            }
+          ]
         end
 
         def determine_status(set_specific_status, build_status = nil, pipeline_build_status = nil)
--- a/app/lib/bk/compat/parsers/bitrise/github_status.rb
+++ b/app/lib/bk/compat/parsers/bitrise/github_status.rb
@@ -6,32 +6,15 @@ module BK
       # Handles the translation of github-status
       class Translator
         def translate_github_status(inputs)
-          context = inputs['context'] || inputs['status_identifier'] || 'continuous-integration/buildkite'
-          warnings = generate_warnings_for_unsupported_options(inputs.keys)
-
-          github_status = {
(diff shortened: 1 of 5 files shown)


app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -5,7 +5,7 @@ module Compat
     module BitriseSteps
       # Implementation of Bitrise step translations
       class Translator
-        VALID_STEP_TYPES = %w[bundler brew-install change-workdir git-clone git-tag script].freeze
+        VALID_STEP_TYPES = %w[bundler brew-install change-workdir git-clone git-tag script github-status].freeze
```
toote, line 8: just nit-picking here, but it would be great to keep the alphabetical order here:

```suggestion
        VALID_STEP_TYPES = %w[bundler brew-install change-workdir git-clone github-status git-tag script].freeze
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -71,6 +71,62 @@ def translate_git_tag(inputs)
...
+            description: inputs['description'] || '',
+            context: inputs['context'] || 'default'
+          }
+        end
+
+        def generate_github_status_script(params)
+          <<~SCRIPT
+            # GitHub Status API call
+            # Extract owner/repo from repository URL
+            #{params[:repo_path_command]}
```
toote, line 99: actually, the extraction of the owner/repo has already been done at this point so you should change the comment or move the call to the `extract_repo_path` function to here

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -71,6 +71,62 @@ def translate_git_tag(inputs)
...
+            auth_token: inputs['auth_token'],
+            api_base_url: inputs['api_base_url'].to_s.chomp('/'),
+            repo_path_command: extract_repo_path(inputs['repository_url']),
+            commit_hash: inputs['commit_hash'],
+            state_command: determine_status(inputs['set_specific_status']),
+            target_url: inputs['pipeline_build_url'] || inputs['build_url'] || '',
+            description: inputs['description'] || '',
+            context: inputs['context'] || 'default'
+          }
+        end
```
toote, line 93: if you are just trying to return a dictionary with good defaults without changing their names:provide good defaults a better way would be to have a hash with the default values and then just call `merge` with the inputs:

```suggestion
          defaults = {
            KEYS: VALUES...
          }
          
          generate_github_status_script(defaults.merge(inputs))
        end
```

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -71,6 +71,62 @@ def translate_git_tag(inputs)
...
+              "$STATUS_API_URL"
+            CURL_RESULT=$?
+            [ $CURL_RESULT -eq 0 ] || echo "Error sending status update to GitHub"
+          SCRIPT
+        end
+
+        def determine_status(set_specific_status)
+          return set_specific_status unless set_specific_status == 'auto'
+
+          "'$([ \"$BUILDKITE_COMMAND_EXIT_STATUS\" = \"0\" ] && echo \"success\" || echo \"failure\")'"
```
toote, line 120: the `BUILDKITE_COMMAND_EXIT_STATUS` variable only exists for hooks after the last command hook. As this script is being added as a command, it wouldn't be set

app/lib/bk/compat/parsers/bitrise/steps.rb
```diff
@@ -71,6 +71,62 @@ def translate_git_tag(inputs)
         def translate_script(inputs)
           inputs['content']
         end
+
+        def translate_github_status(inputs)
+          return '# Invalid github-status step configuration!' unless required_github_status_inputs(inputs)
```
toote, line 76: this should not do a `return` (that would just add the comment to the step's command) but a `raise` instead to signal a huge error in translation of the step

toote (commented):
as the `notify` elements are being added as part of the step's `commands` it makes the generated pipeline invalid. You need to add `notify` as a valid element of a command and then instantiate that for it to be merged with the rest of the command.

toote (changes requested):
while it is a great improvement to what was before, note that:
* the only valid option for `notify`'s `github_commit_status` is `context`
* having multiple notify does not make sense unless they involve different contexts

This is due to the differences in nature from how bitrise and buildkite work

app/lib/bk/compat/parsers/bitrise/github_status.rb
```diff
@@ -0,0 +1,70 @@
...
+        end
+
+        # Command step to update commit status correctly
+        def generate_github_status_command(params)
+          [
+            "buildkite-agent meta-data set github_commit_status_description '#{params[:description]}'",
+            "buildkite-agent meta-data set github_commit_status_status '#{params[:status]}'",
+            "buildkite-agent meta-data set github_commit_status_target_url '#{params[:target_url]}'"
+          ]
+        end
```
toote, line 50: I don't think that these provide any kind of value to the translation. What's more, meta-data is per-build not per step so they are very likely to be overwritten over and over again for no one's benefit.

app/lib/bk/compat/parsers/bitrise/github_status.rb
```diff
@@ -0,0 +1,70 @@
...
+
+          if pipeline_success && build_status == '0'
+            'success'
+          else
+            'failure'
+          end
+        end
+
+        def required_github_status_inputs(inputs)
+          %w[context description set_specific_status].all? { |key| inputs.include?(key) }
```
toote, line 65: I checked the documentation and:
* [`description`](https://github.com/bitrise-io/bitrise-steplib/blob/master/steps/github-status/3.0.1/step.yml#L82-L89) is not mandatory
* [`set_specific_status`](https://github.com/bitrise-io/bitrise-steplib/blob/master/steps/github-status/3.0.1/step.yml#L67-L81) is not mandatory
* `context` does not exist

toote (changes requested):

app/lib/bk/compat/parsers/bitrise/github_status.rb
```diff
@@ -0,0 +1,39 @@
...
+            'context' => context
+          }
+
+          step = BK::Compat::CommandStep.new
+
+          step.notify = [{ 'github_commit_status' => github_status }]
+          step.add_commands(warnings) unless warnings.empty?
+
+          step
+        end
```
toote, line 22: You can create the command object with the corresponding elements instead of assigning them and then invoking the value for returning:

```suggestion
          BK::Compat::CommandStep.new(
            notify: [{ 'github_commit_status' => { 'context' => context } }]
            commands: [warnings].compact
          )
```

toote (changes requested):

toote (approved):

PR #202, opened 2025-03-02
SUP_2437_github_release_translation

phanichowdary9988:
This PR is created to translate github-release for bitrise

### DIFF
--- a/app/examples/bitrise/github-release.yml
+++ b/app/examples/bitrise/github-release.yml
@@ -3,7 +3,7 @@ default_step_lib_source: https://github.com/bitrise-steplib/steps-github-release
 project_type: ios
 
 workflows:
-  build:
+  test-ssh:
     steps:
     - git-clone@8.2.2:
         inputs:
@@ -24,28 +24,28 @@ workflows:
         - draft: no
         - pre_release: no
         - files_to_upload: |
-            /tmp/bitrise-ex/artifacts/app-debug.apk|app-debug.apk
-            /tmp/bitrise-ex/artifacts/app-debug-androidTest.apk|app-debug-androidTest.apk
             /tmp/bitrise-ex/artifacts/no-alt-filename.apk
+            /tmp/bitrise-ex/artifacts/alt-filename.apk|alt-filename.apk
+            /tmp/bitrise-ex/artifacts/alt-filename.apk|alt-different-filename.apk
             /tmp/bitrise-ex/artifacts/${APP_NAME}-debug.apk
-            /tmp/bitrise-ex/artifacts/${APP_NAME}-debug-androidTest.apk
-            /tmp/bitrise-ex/artifacts/app-debug.apk|${APP_NAME}-debug.apk
-            /tmp/bitrise-ex/artifacts/app-debug-androidTest.apk|${APP_NAME}-debug-androidTest.apk
+            /tmp/bitrise-ex/artifacts/alt-filename-${APP_NAME}.apk|alt-filename-${APP_NAME}.apk
+            /tmp/${SUB_PATH}/artifacts/${APP_NAME}-debug.apk|${APP_NAME}-debug.apk
+            ${PATH}/${APP_NAME}-debug.apk|${APP_NAME}-debug.apk
         - api_base_url: https://api.github.com
         - upload_base_url: https://uploads.github.com
         - generate_release_notes: no
 
-  test:
+  test-git:
     steps:
     - git-clone@8.2.2:
         inputs:
         - clone_into_dir: /tmp/bitrise-ex/
-        - repository_url: git@github.com:example/example-repository.git
+        - repository_url: git://github.com/username/repo.git
     - github-release@3.0.1:
         inputs:
         - api_token: $GITHUB_TOKEN
         - username: example
-        - repository_url: git@github.com:example/example-repository.git
+        - repository_url: git://github.com/username/repo.git
         - tag: v1.0.0
         - commit: $BITRISE_GIT_COMMIT
         - name: Release v1.0.0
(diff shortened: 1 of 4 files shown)


app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+            repository_url: inputs['repository_url'],
+            tag: inputs['tag'],
+            body: inputs['body'] || '',
+            draft: inputs['draft'] || 'yes',
+            pre_release: inputs['pre_release'] || 'no',
+            files_to_upload: inputs['files_to_upload'],
+            api_base_url: inputs['api_base_url'] || 'https://api.github.com',
+            upload_base_url: inputs['upload_base_url'] || 'https://uploads.github.com',
+            generate_release_notes: inputs['generate_release_notes'] || 'no'
+          }
```
toote, line 38: A cleaner way to provide the defaults would be so that you don't have to add all the keys that are mandatory:

```suggestion
          defaults = {
            'body' => '',
            'draft' => 'yes',
            'pre_release' => 'no',
            'api_base_url' => 'https://api.github.com',
            'upload_base_url' => 'https://uploads.github.com',
            'generate_release_notes' => 'no'
          }
          
          defaults.merge(inputs)
```

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          owner, repo_name = repo&.split('/') || []
+          raise 'Invalid repository URL!' if owner.nil? || repo_name.nil?
+
+          [owner, repo_name]
+        end
+
+        def build_commands(config, owner, repo_name)
+          commands = []
+          commands << "git clone #{config[:repository_url]} /tmp/bitrise-ex"
```
toote, line 58: this would only be necessary if the repository is different to the one being run in the pipeline, what's more, the examples themselves include a `git-clone` bitrise step that gets translated to a comment stating that it is not necessary, only for it to be done by this plugin immediately afterwards... and, on top of that, the `/tmp/bitrise-ex` being hardcoded doesn't look right

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+          commands << './build-app.sh'
```
toote, line 59: where does this command come from? I don't think that this translation includes the building commands, it just assumes that they were done before... as the examples show that this ended up being duplicated because it was already included in the `script` bitrise step

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+            }
+
+            commands << build_upload_asset_command(upload_config)
+          end
+        end
+
+        def build_upload_asset_command(upload_config)
+          <<~CMD.strip
+            curl -X POST \\
+            -H "Authorization: token #{upload_config[:api_token]}" \\
```
toote, line 87: having a token in commands is normally a security issue but I understand that it is implemented that way in the original bitrise step. I would replace this with a variable name and possibly an assignment with a warning for the user to review at a later step

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+          when %r{^https://}, /^git@/, %r{^ssh://}
+            url = url.sub(%r{^(https://|git@|ssh://)}, '')
+            _, repo = url.split(%r{[:/]}, 2)
+          end
+
+          owner, repo_name = repo&.split('/') || []
+          raise 'Invalid repository URL!' if owner.nil? || repo_name.nil?
+
+          [owner, repo_name]
+        end
```
toote, line 54: this extraction is only done to later be re-joined so I don't think it needs to be so thorough. Usin ruby's builtin `uri` library would also make it a lot more robust and simpler:

```suggestion
        def parse_repo(url)
          data = URI.parse(url.chomp('.git'))
          
          data.path
        end
```

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,97 @@
...
+
+          add_upload_commands(commands, config, owner, repo_name) if config[:files_to_upload]
+
+          commands
+        end
+
+        def add_upload_commands(commands, config, owner, repo_name)
+          config[:files_to_upload].split("\n").each do |file|
+            file_path, file_name = file.split('|')
+            file_name ||= File.basename(file_path)
```
toote, line 69: does `File.basename` work when the path includes a variable that will be escaped by the shell?

toote (changes requested):

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,89 @@
...
+            'generate_release_notes' => 'no'
+          }
+
+          defaults.merge(inputs)
+        end
+
+        def parse_repo(url)
+          url = url.chomp('.git')
+
+          # URI.parse does not handle SSH-style URLs
```
toote, line 44: good catch!

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,89 @@
...
+          path = if url.start_with?('git@')
+                   url.split(':', 2)[1]
+                 else
+                   URI.parse(url).path
+                 end
+
+          path_parts = path.split('/').reject(&:empty?)
+          raise 'Invalid repository URL!' if path_parts.size < 2
+
+          "#{path_parts[-2]}/#{path_parts[-1]}"
```
toote, line 54: I suggested using the `URI.parse` method because valid repository URLs are vast. From what I could test, the main issue with _raw_ git URLs is the lack of a protocol so it may be easier to just add one:

```suggestion
          # URI.parse does not handle SSH-style URLs without protocol
          url = "ssh://#{url}" unless url.index('://')
          
          URI.parse(url).path
```

app/examples/bitrise/github-release.yml
```diff
@@ -0,0 +1,55 @@
...
+        - repository_url: git@github.com:example/example-repository.git
+        - tag: v1.0.0
+        - commit: $BITRISE_GIT_COMMIT
+        - name: Release v1.0.0
+        - body: This is the release body.
+        - draft: no
+        - pre_release: no
+        - files_to_upload: |
+            /tmp/bitrise-ex/artifacts/app-debug.apk|app-debug.apk
+            /tmp/bitrise-ex/artifacts/app-debug-androidTest.apk|app-debug-androidTest.apk
```
toote, line 28: as per the original documentation I would be interested in seeing a couple more test cases:

* when no alternative filename is provided
* when the original filename contains a variable
* when the proposed filename contains a variable

toote (changes requested):

app/lib/bk/compat/parsers/bitrise/github_release.rb
phanichowdary9988: Thanks for the suggestion! I looked into adding "ssh://" when the URL doesn't contain ://, but I ran into an issue with SSH-style Git URLs like git@github.com:user/repo.git. Simply prepending "ssh://" to these URLs results in an invalid format (ssh://git@github.com:user/repo.git).

To ensure compatibility, I updated the logic to handle SSH-style URLs explicitly by replacing the first occurrence of : with / before adding "ssh://". This makes it a valid URI while still supporting other formats correctly. Let me know if you see any concerns with this approach!

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,90 @@
...
+          defaults.merge(inputs)
+        end
+
+        def parse_repo(url)
+          url = url.chomp('.git')
+
+          if url.start_with?('git@')
+            url = url.sub(':', '/')
+            url = "ssh://#{url}"
+          end
```
toote, line 47: I don't think that it is a good idea to check for `git@` because SSH-like URIs may have any user (github uses `git`). I would instead check for the absence of the protocol instead (`://`)

app/examples/bitrise/github-release.yml
toote: this still applies

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,90 @@
...
+
+          uri = URI.parse(url)
+          path_parts = uri.path.split('/').reject(&:empty?)
+
+          raise 'Invalid repository URL!' if path_parts.size < 2
+
+          owner, repo_name = path_parts[-2..]
+          [owner, repo_name]
+        end
```
toote, line 56: you are returning an array here and not a string which means the result looks something like:

```
    https://uploads.github.com/repos/["example", "example-repository"]/releases/assets?name=${APP_NAME}-debug-androidTest.apk
```

You should use the fact that the URI module already parsed it for you:

```suggestion
          uri = URI.parse(url)
          uri.path
        end
```

toote (changes requested):

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        def build_upload_asset_command(upload_config)
+          <<~CMD.strip
+            curl -X POST \\
+            -H "Authorization: token $GITHUB_API_TOKEN" \\
+            -H "Accept: application/vnd.github.v3+json" \\
+            -H "Content-Type: application/octet-stream" \\
+            --data-binary @#{upload_config[:file_path]} \\
+            #{upload_config[:upload_base_url]}/repos/#{upload_config[:repo_path]}/releases/assets?name=#{upload_config[:file_name]}
```
toote, line 81: now that this string can have variables, it may make sense to enclose in in double quotes in case one of them resolves to something with a space

app/lib/bk/compat/parsers/bitrise/github_release.rb
```diff
@@ -0,0 +1,87 @@
...
+
+          if !url.include?('://') && url.include?(':')
+            url = url.sub(':', '/')
+            url = "ssh://#{url}"
+          end
+          path = URI.parse(url).path
+          path_parts = path.split('/').reject(&:empty?)
+          raise 'Invalid repository URL!' if path_parts.size < 2
+
+          "#{path_parts[-2]}/#{path_parts[-1]}"
```
toote, line 52: this should not be nessary... whatever is in the path should be a valid repository

toote (changes requested):

toote (approved):

PR #209, opened 2025-03-21
SUP_2401_SUP_2402

phanichowdary9988:
This PR is created to translate environment variables from Bitrise to Buildkite

### DIFF
--- a/app/examples/bitrise/bitrise.yml
+++ b/app/examples/bitrise/bitrise.yml
@@ -1,0 +1,21 @@
+format_version: "11"
+default_step_lib_source: https://github.com/example/example-bitrise.git
+project_type: other
+
+app:
+  envs:
+  - opts:
+      is_expand: false
+    TEST_KEY: test value
+    
+workflows:
+  deploy-alpha:
+    envs:
+    - ENV_TYPE: alpha
+    - DEPLOY_TARGET: staging
+    steps:
+    - script:
+        inputs:
+        - content: |
+            #!/bin/bash
+            envman add --key MY_RELEASE_NOTE --value "This is the release note for alpha"
--- a/app/examples/bitrise/bitrise.yml
+++ b/app/examples/bitrise/bitrise.yml
@@ -8,9 +8,6 @@ app:
     - APP_ENV: production
       opts:
         is_expand: true
-    - API_KEY: app_level_key
-      opts:
-        is_sensitive: true
 
 workflows:
   primary:
--- a/app/examples/bitrise/bitrise.yml
+++ b/app/examples/bitrise/bitrise.yml
@@ -1,21 +1,49 @@
 format_version: "11"
 default_step_lib_source: https://github.com/example/example-bitrise.git
-project_type: other
+project_type: ios
 
+# App-level environment variables
 app:
   envs:
-  - opts:
-      is_expand: false
-    TEST_KEY: test value
-    
+    - APP_ENV: production
+      opts:
+        is_expand: true
+    - API_KEY: app_level_key
+      opts:
+        is_sensitive: true
+
 workflows:
-  deploy-alpha:
+  primary:
+    # Workflow-level environment variables
     envs:
-    - ENV_TYPE: alpha
-    - DEPLOY_TARGET: staging
+      - WORKFLOW_ENV: staging
+      - BUILD_TYPE: release
+    
     steps:
-    - script:
-        inputs:
-        - content: |
-            #!/bin/bash
-            envman add --key MY_RELEASE_NOTE --value "This is the release note for alpha"
+      - script:
+          title: Skip checkout
+          inputs:
+            - content: "# No need for checkout, the agent takes care of that"
+      
+      - script:
+          title: Install dependencies
+          inputs:
+            - content: pip install -r requirements/dev.txt
+          # Step-level environment variables
+          envs:
+            - PIP_EXTRA_INDEX_URL: https://example.com/pypi
+      
+      - script:
+          title: Set release notes
+          inputs:
+            - content: |
+                #!/bin/bash
+                echo "App ENV: $APP_ENV"
+                echo "Workflow ENV: $WORKFLOW_ENV"
+                echo "Setting MY_RELEASE_NOTE for alpha"
+                export MY_RELEASE_NOTE="This is the release note for alpha"
+                echo "My Release Note: $MY_RELEASE_NOTE"
+          # Step-level environment variables
+          envs:
+            - ENV_TYPE: alpha
+            - DEPLOY_TARGET: staging
--- a/app/lib/bk/compat/models/pipeline.rb
+++ b/app/lib/bk/compat/models/pipeline.rb
@@ -29,7 +29,8 @@ module BK
 
       def to_h
         {}.tap do |h|
-          h[:env] = @env unless @env.empty?
+          # Conditionally include the env section only if it contains valid environment variables
+          h[:env] = @env unless @env.empty? || @env.keys.all? { |k| k.to_s.strip.empty? }
           h[:steps] = simplify_steps
           h[:agents] = @agents unless @agents.empty?
         end
--- a/app/lib/bk/compat/models/pipeline.rb
+++ b/app/lib/bk/compat/models/pipeline.rb
@@ -29,7 +29,6 @@ module BK
 
       def to_h
         {}.tap do |h|
-          # Conditionally include the env section only if it contains valid environment variables
           h[:env] = @env unless @env.empty?
           h[:steps] = simplify_steps
           h[:agents] = @agents unless @agents.empty?
--- a/app/lib/bk/compat/models/pipeline.rb
+++ b/app/lib/bk/compat/models/pipeline.rb
@@ -30,7 +30,7 @@ module BK
       def to_h
         {}.tap do |h|
           # Conditionally include the env section only if it contains valid environment variables
-          h[:env] = @env unless @env.empty? || @env.keys.all? { |k| k.to_s.strip.empty? }
+          h[:env] = @env unless @env.empty?
           h[:steps] = simplify_steps
           h[:agents] = @agents unless @agents.empty?
         end
--- a/app/lib/bk/compat/models/steps/command.rb
+++ b/app/lib/bk/compat/models/steps/command.rb
@@ -85,10 +85,6 @@ module BK
         list_attributes.each { |a| send(a).concat(other.send(a)) }
         hash_attributes.each { |a| send(a).merge!(other.send(a)) }
 
-        # Merge environment variables
-        @env ||= {}
-        @env.merge!(other.env) if other.respond_to?(:env)
-
         update_attributes!(other)
       end
 
@@ -96,10 +92,6 @@ module BK
         # almost the same as merge but self/other are reversed here
         list_attributes.each { |a| send("#{a}=", other.send(a).concat(send(a))) }
         hash_attributes.each { |a| send("#{a}=", other.send(a).merge(send(a))) }
-
-        # Merge environment variables
-        @env ||= {}
-        @env.merge!(other.env) if other.respond_to?(:env)
 
         update_attributes!(other)
       end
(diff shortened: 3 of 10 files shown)


app/lib/bk/compat/models/pipeline.rb
```diff
@@ -29,7 +29,8 @@ def simplify_steps
 
       def to_h
         {}.tap do |h|
-          h[:env] = @env unless @env.empty?
+          # Conditionally include the env section only if it contains valid environment variables
+          h[:env] = @env unless @env.empty? || @env.keys.all? { |k| k.to_s.strip.empty? }
```
toote, line 33: I don't think that this change should be necessary, the checks for that should be done before adding the values to the hash

app/lib/bk/compat/models/steps/command.rb
```diff
@@ -85,6 +85,10 @@ def merge!(other)
         list_attributes.each { |a| send(a).concat(other.send(a)) }
         hash_attributes.each { |a| send(a).merge!(other.send(a)) }
 
+        # Merge environment variables
+        @env ||= {}
+        @env.merge!(other.env) if other.respond_to?(:env)
+
```
toote, line 91: this should already been taken care of as `@env` is part of `hash_attributes`

app/lib/bk/compat/models/steps/command.rb
```diff
@@ -93,6 +97,10 @@ def pre_merge!(other)
         list_attributes.each { |a| send("#{a}=", other.send(a).concat(send(a))) }
         hash_attributes.each { |a| send("#{a}=", other.send(a).merge(send(a))) }
 
+        # Merge environment variables
+        @env ||= {}
+        @env.merge!(other.env) if other.respond_to?(:env)
+
```
toote, line 103: this should already been taken care of as `@env` is part of `hash_attributes`

app/lib/bk/compat/parsers/bitrise.rb
```diff
@@ -41,12 +41,29 @@ def parse
...
         stack_value = @config.dig('meta', 'bitrise.io', 'stack')
         agents = stack_value ? { stack: stack_value } : {}
+
+        env = process_app_envs
+
         Pipeline.new(
           steps: bk_steps,
-          agents: agents
+          agents: agents,
+          env: env
```
toote, line 50: why not call the method directly instead of creating an intermediate variable?

```suggestion
        Pipeline.new(
          steps: bk_steps,
          agents: agents,
          env: process_app_envs
```

app/lib/bk/compat/parsers/bitrise.rb
```diff
@@ -41,12 +41,29 @@ def parse
...
+          env_var.each do |key, value|
+            next if key == 'opts'
+            next if key.to_s.strip.empty? || value.to_s.strip.empty?
+
+            env[key] = value
+          end
+        end
+        env
+      end
+
```
toote, line 66: It is complicated to follow the logic of this function that just maps the elements removing any `opts` if there are. A simpler version would be:

```suggestion
      def process_app_envs(vars)
        return {} unless vars # takes care of empty vars and nils
        
        envs.map do |env_var|
           env_var.delete('opts')
           
           # if there is more than one thing, it is invalid
           raise unless env_var.one?
           
           env_var.to_a
         end.to_h
      end

```

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -14,10 +14,25 @@ def parse_workflow(wf_name, wf_config)
...
+        end
+      end
+
+      def process_env_variable(env, target_env)
+        env.each do |key, value|
+          next if key == 'opts' || key.to_s.strip.empty? || value.to_s.strip.empty?
+
+          target_env[key] = value
+        end
+      end
```
toote, line 34: 1- this is the same method as in the `bitrise.rb`, it should be moved to a different file or a helper to avoid duplication
2- I have provided an alternative version that would generalize it enough to be used in both places

toote (changes requested):
in addition to the other comments, according to the [documentation](https://devcenter.bitrise.io/en/builds/environment-variables.html#setting-env-vars-in-the-bitrise-yml) and the [schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/bitrise.json#L823), steps can also accept the `envs` and that case is not tested nor handled (and makes a more compelling case for modularizing the processing of environment variables)

app/examples/bitrise/bitrise.yml
```diff
@@ -0,0 +1,49 @@
...
+
+# App-level environment variables
+app:
+  envs:
+    - APP_ENV: production
+      opts:
+        is_expand: true
+    - API_KEY: app_level_key
+      opts:
+        is_sensitive: true
```
toote, line 13: I couldn't find any reference to this opt anywhere, do you have any documentation or link where it explains the usage of this?

app/lib/bk/compat/parsers/bitrise.rb
```diff
@@ -41,12 +42,22 @@ def parse
...
       end
 
+      def process_app_envs(vars)
+        return {} unless vars # takes care of empty vars and nils
+
+        result = {}
+        BK::Compat::Bitrise::EnvHelper.process_env_variables(vars, result)
+        result
+      end
+
```
toote, line 60: This should just be a call to `BK::Compat::Bitrise::EnvHelper.process_env_variables` that just returns the corresponding dict

app/lib/bk/compat/parsers/bitrise/env_helper.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module BK
+  module Compat
+    class Bitrise
+      # Helper methods for processing environment variables in Bitrise configuration
+      module EnvHelper
+        def self.process_env_variable(env, target_env)
```
toote, line 8: why take `target_env` as a parameter. You should just return the resulting dictionary instead

app/lib/bk/compat/parsers/bitrise/env_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+          return unless env.is_a?(Hash)
+
+          env.each do |key, value|
+            next if key == 'opts' || key.to_s.strip.empty? || value.to_s.strip.empty?
+
+            target_env[key] = value
+          end
```
toote, line 15: an easier way to do this is:

```suggestion
          env.delete!('opts')
          env
```

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -14,20 +16,43 @@ def parse_workflow(wf_name, wf_config)
           step_key, step_config = step.first
           cmd_step << translate_step(step_key, step_config)
         end
-
+        process_workflow_envs(wf_config, cmd_step)
         cmd_step
       end
 
+      def process_workflow_envs(wf_config, cmd_step)
+        cmd_step.env ||= {}
```
toote, line 24: commands class already take care of this, there is no need to ensure it is a hash that is empty

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -14,20 +16,43 @@ def parse_workflow(wf_name, wf_config)
...
+        BK::Compat::Bitrise::EnvHelper.process_env_variables(wf_config['envs'], cmd_step.env)
```
toote, line 25: if `BK::Compat::Bitrise::EnvHelper.process_env_variables` didn't take the second parameter and returned a dictionary (as mentioned in other comments) this should be:

```suggestion
        cmd_step.env.merge!(BK::Compat::Bitrise::EnvHelper.process_env_variables(wf_config['envs']))
```

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -14,20 +16,43 @@ def parse_workflow(wf_name, wf_config)
...
+
+          # Apply environment variables if available
+          apply_env_config(translated_step, step_config)
         end
       end
 
+      def apply_env_config(step, config)
+        return unless step.respond_to?(:env) && config['envs']
+
+        step.env ||= {}
```
toote, line 52: the command class already takes care of this so it shouldn't be necessary

app/lib/bk/compat/parsers/bitrise/workflows.rb
```diff
@@ -14,20 +16,43 @@ def parse_workflow(wf_name, wf_config)
...
+        BK::Compat::Bitrise::EnvHelper.process_env_variables(config['envs'], step.env)
```
toote, line 53: if `BK::Compat::Bitrise::EnvHelper.process_env_variables` didn't take the second parameter and returned a dictionary (as mentioned in other comments) this should be:

```suggestion
        step.env.merge!(BK::Compat::Bitrise::EnvHelper.process_env_variables(config['envs']))
```

toote (changes requested):

app/examples/bitrise/bitrise.yml
phanichowdary9988: Here is the documentation:
https://devcenter.bitrise.io/en/builds/environment-variables.html#replacing-variables-in-inputs

app/examples/bitrise/bitrise.yml
```diff
@@ -0,0 +1,49 @@
...
+                #!/bin/bash
+                echo "App ENV: $APP_ENV"
+                echo "Workflow ENV: $WORKFLOW_ENV"
+                echo "Setting MY_RELEASE_NOTE for alpha"
+                export MY_RELEASE_NOTE="This is the release note for alpha"
+                echo "My Release Note: $MY_RELEASE_NOTE"
+          # Step-level environment variables
+          envs:
+            - ENV_TYPE: alpha
+            - DEPLOY_TARGET: staging
```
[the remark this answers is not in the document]
phanichowdary9988: From bitrise documentation(https://devcenter.bitrise.io/en/builds/environment-variables.html) ,only pipeline and workflow level envs allowed. I have deleted those envs from yaml file

app/examples/bitrise/bitrise.yml
toote: I couldn't find any reference to `is_sensitive` in that page

app/examples/bitrise/bitrise.yml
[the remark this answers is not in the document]
toote: In that same page, under the heading [Scope of Environment Variables](https://devcenter.bitrise.io/en/builds/environment-variables.html#scope-of-environment-variables) it states that step level variables are valid

app/lib/bk/compat/models/steps/command.rb
toote: for a third time, this change should not be necessary

app/lib/bk/compat/models/steps/command.rb
toote: as mentioned before, this change should not be necessary

app/lib/bk/compat/parsers/bitrise.rb
```diff
@@ -41,18 +42,67 @@ def parse
...
+          env: process_app_envs(@config.dig('app', 'envs') || {})
         )
       end
 
+      def process_app_envs(vars)
+        return {} unless vars
+
+        BK::Compat::Bitrise::EnvHelper.process_env_variables(vars)
+      end
+
```
toote, line 58: `BK::Compat::Bitrise::EnvHelper.process_env_variables` already takes care of empty arrays/hashes and `nil`s so this should be simplified to:

```suggestion
          env: BK::Compat::Bitrise::EnvHelper.process_env_variables(@config.dig('app', 'envs')
      end
```

app/lib/bk/compat/parsers/bitrise.rb
```diff
@@ -41,18 +42,67 @@ def parse
...
+        return unless step.respond_to?(:env) && config['envs']
+
+        step.env.merge!(BK::Compat::Bitrise::EnvHelper.process_env_variables(config['envs']))
+      end
+
+      def load_workflow_step(step_key, step_inputs)
+        step_key_normalized = step_key.split('@').first
+        # Translate step with normalized key (for translator) and its input configuration
+        @translator.translate_step(step_key_normalized, step_inputs)
+      end
```
toote, line 105: why were these methods copied here from the workflow file?

toote (changes requested):

toote (approved):

PR #217, opened 2025-05-11
Added bug repot and feature request templates

james2791:
Added standard templates for filing:
- Bug and/or support requests
- Feature requests (new parameter translation, new provider support etc)


mcncl:
@james2791 I don't know if the `vendor requests` file is needed? I have a feeling that folks will raise additional support as feature requests. Traditionally we'd use a bug report and a feature request for templates.

james2791:
> @james2791 I don't know if the `vendor requests` file is needed? I have a feeling that folks will raise additional support as feature requests. Traditionally we'd use a bug report and a feature request for templates.
> 
> Just my 2c.

Agree @mcncl, pushed some alts to put all things relevant to features/new provider support/params to the FR template, support one stays as is 👍

CoreyGinnivan:
@mcncl Yeah fair enough! I slapped those ones together because I was gonna link to them directly but those links also might change too - but regardless think these templates could be cool, so let's do ittt

CoreyGinnivan (approved):

mcncl (approved):

PR #226, opened 2025-05-22
Jenkins parameters

toote:
Very crude translation using a Regexp... not ideal but I remember the pain it was to create the parser for GHA so going this way as a first approach for the time being

Note: merge #225 before this


james2791 (approved):
💨

PR #233, opened 2025-06-20
Jenkins `options` translation

toote:
This was a little more convoluted than I expected. I also found a few missing keys for the `Command` models as well as a bug or two


james2791 (approved):
💨 💪

PR #249, opened 2025-09-04
Add Syntax Error Handling On YAML Format Issues

lizrabuya:
-  Return an error message when a YAML formatting issue is encountered instead of having the tool crash. This includes error handling of `Psych::SyntaxError` exceptions. 
- Increased MethodLength limit to cater to the added error handling of exceptions in code.  Additional lines in code to catch the exception has caused the circleci `self.matches?` method exceed the MethodLength in `rubocop` and returned a Lint Error.

### DIFF
--- a/app/lib/bk/compat/parsers/circleci.rb
+++ b/app/lib/bk/compat/parsers/circleci.rb
@@ -46,7 +46,7 @@ module BK
           config.fetch('version', 2.1) == 2 && config['jobs'].include?('build')
         end
       rescue Psych::SyntaxError
-        return false
+        false
       end
 
       def initialize(text, options = {})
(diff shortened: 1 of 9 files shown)


app/lib/bk/compat/parsers/circleci.rb
```diff
@@ -45,16 +45,20 @@ def self.matches?(text)
...
       def initialize(text, options = {})
-        @config = YAML.safe_load(text, aliases: true)
-        @now = Time.now
-        @options = options
-        @commands_by_key = {}
-        @executors = {}
-
-        register_translators!
+        BK::Compat::Error::CompatError.safe_yaml do
+          @config = YAML.safe_load(text, aliases: true)
```
toote, line 54: only this line should be part of the block and (as mentioned in other places) if you create a helper method, it would be a lot less change:

```suggestion
          @config = MODULE::load_yaml(text, aliases: true)
```

toote (changes requested):
The idea is really good, but I think there are a lot of issues with the implementation itself. Happy to pair up and work on this!

